Issue #1, opened 2016-05-30
Linux wheels


c-martinez:
I saw there is already a ubuntu wheel, but came across this, and thought it might be useful:

https://github.com/pypa/manylinux

skvark:
Looks interesting. I will definetely at least test if I can get the manylinux builds working (would be great for other than Ubuntu users).

The current builds for Windows and Ubuntu are now passing, but I haven't pushed the results yet to PyPI (setup.py is incomplete and I have to do some cleaning). Next phase is to get OS X builds working in Travis, and after that I will propably test the manylinux thing.

PR #2, opened 2016-06-30
Build on Travis CI using manylinux

c-martinez:
This PR should address previous suggestion #1. Builds on Travis seem fine (except for 2.6).

Couple of issues
 - Currently not building for 2.6, because it complains gives `ImportError: cannot import name SkipTest`
 - `build-wheels.sh` will need to be changed in the future:
    - It force-updates `auditwheel` (until the docker image updates auditwheel)
    - It clones `OpenCV` from [my own fork](https://github.com/c-martinez/opencv.git) -- if this is the way forward to build OpenCV wheels, I'll send a PR upstream to incorporate the few changes required.

Is this useful?


skvark:
Appveyor build file needs minor fix, but otherwise LGTM.

I don't really know the answer to the question if this is the way forward to build OpenCV wheels. If you have a better idea, please share. As the OpenCV maintainers are a bit reluctant to include Python packaging support directly to the main repo (https://github.com/Itseez/opencv/issues/6516), I think this separate repository + cloning OpenCV is pretty much the only option there is.

c-martinez:
Changed the Appveyor build (hopefully it works correctly).

I think you are right and cloning OpenCV to build the wheels might be the only option. I guess what we are missing now is registering the wheels in PyPI. Do you have any experience with it? Otherwise I guess I can figure it out.

skvark:
I have experience running a local PyPI and I can do the uploading. However, there are still a few things to do:

* the builds (cloning) have to be done from the main OpenCV repository
* versioning:
    * do we pull only the latest tag and make builds for it or do we also offer dev builds which are essentially built always when there's a commit in the OpenCV main development branch
* OS X wheels
* Are the Linux wheels allowed in PyPI (at some point they weren't)

I will create separate issues for these.

Issue #5, opened 2016-07-04
Check that PyPI supports manylinux wheels


skvark:
Just a minor check.

c-martinez:
[This note](https://packaging.python.org/distributing/#platform-wheels) seems to imply that PyPI does not support platform wheels. But more recently, [PEP513](https://www.python.org/dev/peps/pep-0513/#pypi-support) mentions PyPI support for manylinux wheels. 

More importantly, there are packages in PyPI (such as [Pillow](https://pypi.python.org/pypi/Pillow/3.3.0)) which already provide manylinux wheels. So I think we are ok.

Also, the way the [Pillow wheels](https://github.com/python-pillow/pillow-wheels) are built seems interesting.

skvark:
Yes, it seems that manylinux wheels can be added to PyPi.

The OS X builds could be done as they are done in the Pillow repo. I think also that we could use the same submodule method for obtaining OpenCV. That way the version could be locked to certain commit and updated manually whenever needed.

PR #8, opened 2016-08-07
Build using submodules

c-martinez:
As discussed in #5 , Made a few changes to build using OpenCV as a submodule, and using multibuild (as Pillow does).

OSX build still work in progress.


skvark:
Thanks, I merged this manually.

PR #15, opened 2016-09-21
Copy (ffmpeg) .dll to fix #6

### LINKED ISSUE #6: Pack opencv_ffmpeg310.dll
Build process downloads ffmpeg .dll, but fails to pack it into resulting wheel.

https://github.com/opencv/opencv/pull/6803

techtonik:
This attempts to copy all .dll files in build directory,
where should be only ffmpeg .dll


skvark:
Note that you have to add rule for copying *.dll in setup.py too so that the files are actually copied inside wheels:

https://github.com/skvark/opencv-python/blob/master/setup.py#L33

techtonik:
Thanks for the pointer.

techtonik:
File roller keeps crashing on my Fedora, so I am unable to inspect wheel contents.

techtonik:
Need some sanity test for AppVeyor to ensure that it works (no Windows systems right now there).

Also need to stuff `ffmpeg` license somewhere..

skvark:
You can add the license to this file in same manner as the BSD license has been added (with a note that cv2 folder contains the FFmpeg dll): https://github.com/skvark/opencv-python/blob/master/LICENSE-3RD-PARTY.txt

And it could be useful add a unit test for video reading. If there's e.g. some short mp4 public domain clip it could be added to the test folder and then run simple unit test:

    import cv2
    
    cap = cv2.videoCapture("filename.mp4")
    self.assertTrue(cap.isOpened())

skvark:
Wheels contain now both dlls, wildcard pattern needs to be modifed.

cv2

techtonik:
I used Chromium test files when I discovered these issues - https://chromium.googlesource.com/chromium/src/+/master/media/test/data/

techtonik:
Seems like I am out of time to finish this, so feel free to put the final touches here.

skvark:
I did some testing, the video files open fine if the cv2 folder is in Windows PATH. I'll try to solve this...

techtonik:
@skvark thanks! I am still unable to get more time for OpenCV.

Now people will have less trouble trying to follow https://www.youtube.com/watch?v=MC00XWdl-ms =)

PR #19, opened 2016-12-28
Refactored AppVeyor script

glitchassassin:
Added support for Python 3.4 and 3.4 x64. Split each version into its own AppVeyor job with matrix feature.


skvark:
LGTM. The Appveyor failure seems to be related to new auditwheel version and wheel file structure, I'll fix that before releasing to PyPI.

PR #25, opened 2017-02-05
Ffmpeg osx

cancan101:
This might fix https://github.com/skvark/opencv-python/issues/14 on OS X for Python 2.x by using the brew install for opencv which links against ffmpeg.

### DIFF
--- a/travis/build-wheels-osx.sh
+++ b/travis/build-wheels-osx.sh
@@ -9,6 +9,8 @@ echo 'PYTHON_VERSION: '$PYTHON_VERSION
 echo 'PIP and brew installs'
 
 pip install -r requirements.txt
+brew install ffmpeg
+ffmpeg -L
 
 echo 'Config make'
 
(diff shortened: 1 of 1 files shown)


skvark:
I'm sorry but I can't accept this pull request. Installing OpenCV via homebrew hides the actual build process. Some main issues:

- the build process must be consistent on all Python versions and on all platforms
    - -> cmake and given flags must be visible so that possible issues can be debugged
- build output is not visible

Additionally some tests should be run to make sure that the video functionality works. I'm not sure if OpenCV requires the exactly same FFmpeg version as it was built with in the end user system. If that's the case, it gets more complicated to ship a wheel which works directly out of the box.

cancan101:
That's fine, I consider this more a PoC. Question is whether there are any flags used in the homebrew build that can be pulled over to your current build process to cause opencv to link against ffmpeg, etc.

skvark:
Yes, I'll have to check that. I hope I have soon enough spare time to look at this FFmpeg issue. I'll leave this PR open for future reference.

PR #30, opened 2017-03-27
Relax numpy version requirements

adrianmoisey:
Just want to see if tests pass with this


skvark:
I agree that current implementation is not very good. See my comment in issue https://github.com/skvark/opencv-python/issues/29 how this should be implemented.

skvark:
FYI the Travis failure related to test dependencies on the Ubuntu container has been fixed in https://github.com/skvark/opencv-python/commit/f5b356293b58afc63fe3764521caa4cd8b5a2345.

adrianmoisey:
Thanks. I've cherry picked it across.
Once I get the CI builds passing, I'll cleanup the commit history,

skvark:
Note the numpy requirement in ``requirements.txt`` installs the latest numpy and it overrides the travis.yml numpy version. The requirement should be removed. ``requirements.txt`` is used in these lines and it should be safe to remove these calls completely:

- https://github.com/skvark/opencv-python/blob/master/travis/build-wheels.sh#L14
- https://github.com/skvark/opencv-python/blob/master/travis/build-wheels-osx.sh#L11

For Appveyor I suggest to define a variable which contains the exact numpy version requirement for each build in the matrix and add a new install line for it, e.g. ``"%PYTHON%/python.exe" -m pip install numpy==%NP_VERSION%`` after this line https://github.com/skvark/opencv-python/blob/master/appveyor.yml#L51

adrianmoisey:
Ah right, thanks!
I hadn't looked much at the build scripts.

adrianmoisey:
I'm finished here. What do you think?
If you're happy with this, I'll squash the commits and fix the git history, as I made a few debug commits.

skvark:
Do you want to clean up the commits before the merge?

adrianmoisey:
Yeah, let me do that.
Should I do that in a new branch?
Or should I destroy the history here?

skvark:
Hmm, I think it's easiest if I do the squashing directly from Github UI unless you want to do it manually with git.

adrianmoisey:
That works for me :+1:

skvark:
Thanks!

adrianmoisey:
Thanks for all the help!

PR #32, opened 2017-04-09
Fix name export

levzhazeschi:
Some IDEs (especially PyCharm) are not able to index module contents if `__init__.py` exports them in ways other than this (http://stackoverflow.com/questions/34365044/pycharm-does-not-recognize-cv2-as-a-module). This commit fixes the problem.


levzhazeschi:
It leaves the `cv2` module with only the contents of old `cv`.

Issue #34, opened 2017-05-02
OSX and sudo install requirements


glitchassassin:
When trying to install `opencv-python` on OS X (python3 installed via Homebrew), installing as an unprivileged user failed with the following message:

    Exception:
    Traceback (most recent call last):
    File "/usr/local/lib/python3.6/site-packages/pip/basecommand.py", line 215, in main
        status = self.run(options, args)
    File "/usr/local/lib/python3.6/site-packages/pip/commands/install.py", line 342, in run
        prefix=options.prefix_path,
    File "/usr/local/lib/python3.6/site-packages/pip/req/req_set.py", line 784, in install
        **kwargs
    File "/usr/local/lib/python3.6/site-packages/pip/req/req_install.py", line 851, in install
        self.move_wheel_files(self.source_dir, root=root, prefix=prefix)
    File "/usr/local/lib/python3.6/site-packages/pip/req/req_install.py", line 1064, in move_wheel_files
        isolated=self.isolated,
    File "/usr/local/lib/python3.6/site-packages/pip/wheel.py", line 377, in move_wheel_files
        clobber(source, dest, False, fixer=fixer, filter=filter)
    File "/usr/local/lib/python3.6/site-packages/pip/wheel.py", line 323, in clobber
        shutil.copyfile(srcfile, destfile)
    File "/usr/local/Cellar/python3/3.6.1/Frameworks/Python.framework/Versions/3.6/lib/python3.6/shutil.py", line 121, in copyfile
        with open(dst, 'wb') as fdst:
    PermissionError: [Errno 13] Permission denied: '/usr/local/LICENSE-3RD-PARTY.txt'

After installing with `sudo -H pip3 install opencv-python`, the package worked.

Can the sudo requirement be eliminated by not writing the license to `/usr/local/`?

skvark:
You can try ``pip install --user opencv-python`` or use virtualenvs. And you are correct, sudo should not be used to install packages with pip. 

However, the root of this issue lies in MacOS or some of its package managers handle Python installations. I'm not too famialiar with all the MacOS weirdness, but the ``data_files`` parameter in ``setup.py`` uses relative paths so ``distutils`` behaves  [just like documented](https://docs.python.org/3.6/distutils/setupscript.html#installing-additional-files). If I move the licenses to a subfolder in this package, the issue stays the same. Even if there are no ``data_files``, the problem would still most likely persist.

[...]

PR #41, opened 2017-07-17
License file fix for #34

glitchassassin:
Moves the license files to package_data instead of data_files


skvark:
I think it would be better that the licenses stay in the root of this repository. By moving them to cv2 folder Github can't autodetect the license and it's harder to find them when they are "hidden".

No big changes needed, just copy the licenses to the cv2 folder during build process. This has to be done before setup.py is called.

And don't worry about the broken builds, I'm looking into that issue.

glitchassassin:
Updated the travis yml file to copy those license files as part of the `before_install` step - does this work?

skvark:
Yes, that should do it. Could you also add it to the Windows builds? New xcopy line to appveyor/build.cmd should be enough.

glitchassassin:
Of course, forgot that setup.py needs it there either way. 🙄 My bad.

skvark:
No problem and thanks :) I will merge this when I get the builds fixed.

PR #54, opened 2017-11-14
Adjust build environment to upstread recommendations and to work in XP in Python versions that support it

native-api:
Upstream requires vs11 minimum (I confirmed it to not build in vs90 due to lack of language features) and Py3.5+ uses vs14.

See https://docs.opencv.org/3.1.0/d5/de5/tutorial_py_setup_in_windows.html and https://wiki.python.org/moin/WindowsCompilers#Which_Microsoft_Visual_C.2B-.2B-_compiler_to_use_with_a_specific_Python_version_.3F


skvark:
What's the reasoning behind this? There are no Visual C++ runtime related issues with the current builds as far as I know because OpenCV build is static.

native-api:
The problem is that the 2.7 package doesn't work in XP (I didn't check the 3.4 one but expect as much, too).

I changed the environment to vs11 as well to be as close to those Pythons' runtimes as possible but if you're sure it doesn't matter, feel free to change it back.

skvark:
I'm sorry, I totally missed that XP in the title. Windows XP is an ancient operating system and it's true that software which has been built with VS 2015 does not work in XP (installing Visual C++ 2015 redistributable *might* fix the problem: https://www.microsoft.com/en-US/download/details.aspx?id=48145). 

The old build toolchain is going to add dependency to the older Visual C++ Runtime which means that all existing installations would most likely break during package update if the older runtime isn't present. In addition, Windows XP's market share is approaching slowly 0 % (https://www.neowin.net/news/windows-10-market-share-continues-to-increase-as-xp-falls-below-7-for-the-first-time). 

I hope you understand that I'll have to reject this pull request since Windows XP support is not in the scope of this project.

native-api:
If using v140_xp build toolset -- which is what I suggested by "feel free to change it back", -- it should be compiled against the same runtime as the default thus not introduce any new requirements. I'm checking this as we speak.

It's rather hard to do the work without a script to build locally. Is it okay if I submit one and make appveyor's script use it?

skvark:
Sure, that's okay. The possibility to do a local build is one of the things which should be done anyway at some point since ``setup.py install`` can't currently be used to run the whole build process. And if the toolset does not introduce any new requirements it's ok to make these changes. It's just that the distributions generated via this project are used by a lot of people (more than 2 000 000 downloads already which means that these packages are installed about 4000 times a day) so it's important that the packages just work even after upgrades.

PR #55, opened 2017-11-23
PATH editing only needed in Windows

native-api:
AFAICS, FFMpeg DLL is only bundled in the Windows version.

### DIFF
--- a/cv2/__init__.py
+++ b/cv2/__init__.py
@@ -1,10 +1,4 @@
-import sys
-import os
 import importlib
-
-# FFmpeg dll is not found on Windows without this
-if os.name == 'nt':
-    os.environ["PATH"] += os.pathsep + os.path.dirname(os.path.realpath(__file__))
 
 # make IDE's (PyCharm) autocompletion happy
 from .cv2 import *
--- a/cv2/__init__.py
+++ b/cv2/__init__.py
@@ -3,7 +3,8 @@ import os
 import importlib
 
 # FFmpeg dll is not found on Windows without this
-os.environ["PATH"] += os.pathsep + os.path.dirname(os.path.realpath(__file__))
+if os.name == 'nt':
+    os.environ["PATH"] += os.pathsep + os.path.dirname(os.path.realpath(__file__))
 
 # make IDE's (PyCharm) autocompletion happy
 from .cv2 import *


skvark:
Thanks! However, I'm not sure if that PATH modification is needed anymore (and Python might not be capable to inject that new PATH to the environment of the running process). The structure of the packages has changed after that a lot and it would be good to do a test run without that line and see if tests pass.

native-api:
https://docs.python.org/3/library/os.html?highlight=os%20environ#os.environ:

> If the platform supports the `putenv()` function, this mapping may be used to modify the environment

I can also confirm that `opencv_ffmpeg330_64.dll` library load works in the 3.3.0 release even without `PATH` modification -- yet `PATH` is explored first when looking for it.

native-api:
At this rate, the entire wrapper `__init__.py` stops being necessary. PyCharm is already smart enough to autocomplete symbols from an extension module.

skvark:
I think it's safe to remove that PATH line. ``__init__.py`` must be kept since it's the only way for Python to detect that the folder is actually a package and keep the import interface (``import cv2``) backwards compatible. Without those couple of lines the import would have to be ``from cv2 import cv2`` and this repository would be full of issues like "can't import cv2" because most of the users don't bother to read any documentation :)

Installing to the root of site-packages is a no go (which is the way how OpenCV installation guides usually do it). That's probably not even possible without manual copying.

native-api:
> |__init__.py| must be kept since it's the only way for Python to 
> detect that the folder is actually a package and keep the import 
> interface (|import cv2|) backwards compatible.
>
I mean, ditch the directory altogether and ship just the .pyd (and 
ffmpeg .dll in Windows). I mean, this is like opencv builds the module. 
I reckon the only reason why package and stuff was added is to work 
around problems with the dependent DLL and IDEs.
> Installing to the root of site-packages is a no go (which is the way 
> how OpenCV installation guides usually do it). That's probably not 
> even possible without manual copying. 
Not only is this possible, it's the standard way for standalone modules 
(i.e. those not under a Python package). From what I have here, 
`scandir' (an IPython dependency) is a module like this that has a .pyd .

The other standard way is to place files into a subdirectory and add a 
.pth file to `site-packages' root (it's a list of relative paths to add 
to sys.path, parsed by site.py at startup, see 
https://stackoverflow.com/questions/46969136/can-i-zip-all-the-python-standard-libs-and-the-python-still-able-to-import-it/46970267#46970267 
). I never actually did this but since this is the standard way, it must 
be supported by the stock setup machinery.

-- 
Regards,
Ivan

skvark:
You are correct, it's possible to distribute standalone modules but I'm not willing to ditch the cv2 directory. OpenCV installs the DLL files to the root, but it doesn't mean that it's good practice. Afaik OpenCV's installer just copies the files into the site-packages. There's no package metadata which means that for example the cv2.pyd and its dependencies can't be uninstalled via pip.  

The main reason why I added the cv2 folder when I started developing this project was that it serves as a namespace and isolates the package contents in a more convenient way. This of course doesn't still take into account the fact that there might be a conflicting manually installed OpenCV in the user's site-packages, but at least this package will install/uninstall cleanly without issues in that case. Also one thing to consider is that dropping the namespace at this point would most likely break all existing installations during upgrade. 

IDE support (PyCharm) is related to the namespace in a different way since the import logic in the ``__init__.py`` was originally flawed (PyCharm did not pick up the cv2 module, see this commit: https://github.com/skvark/opencv-python/commit/3140bce97d3415adf0c5e195421d4fec49152849). This was fixed in a recent release.

native-api:
Anyway, I removed the PATH line entirely. Guess we can wrap up this PR now.

skvark:
Thanks!

PR #56, opened 2017-12-07
MacOS FFMpeg builds. 

techdragon:


skvark:
It seems that Travis still has the smaller timeout in use. I'll check this with Travis.

skvark:
The increased build timeout is working now. Before I accept this PR could you enable the video file test in tests/test.py just to be sure that video capture works as expected?

Thank you, I know you have done work also at homebrew side to make this happen.

PR #57, opened 2017-12-13
 Update setup.py file to flag compatibility with Python 2.7-3.4-3.5-3.6

DEKHTIARJonathan:
The PyPI API does not flag OpenCV-Python to be compatible with Python 3 and which version of Python is supported.

The following flags should fix this issue. 
This is based on the wheel files available on: https://pypi.python.org/pypi/opencv-python


skvark:
Thanks, I'll merge this after the builds are done.

PR #108, opened 2018-07-16
add PyPi download count badge to README

scivision:
Thanks for this highly useful package--several million downloaders agree!


skvark:
Thanks!

PR #110, opened 2018-07-19
Fix crash while building opencv for OSX

yonip:
The original crash was caused by scikit-build both prioritizes command line options over the parameters passed to the `setup` function, and that it defaults the command line options for `-DCMAKE_OSX_DEPLOYMENT_TARGET` and for `-DCMAKE_OSX_ARCHITECTURES` without checking that `setup` was passed those parameters. 
([this](https://github.com/scikit-build/scikit-build/blob/d8e4db57e0a930db75bd204e47b64ba83a7300bb/skbuild/setuptools_wrap.py#L468) is where scikit-build defaults the command line parameters, and [this](https://github.com/scikit-build/scikit-build/blob/d8e4db57e0a930db75bd204e47b64ba83a7300bb/skbuild/setuptools_wrap.py#L506) is where the command line parameters are combined with the passed in parameters).
As a workaround, I overrode one function in multibuild's `common_utils.sh` in `config.sh` that allows for extra command line parameters to be passed to setup.py, and set the parameters only for OSX tests in `.travis.yml`.
Also fix a later crash where `install_run` crashed on OSX builds with Python 3.7, where the `TEST_DEPENDS` environment variable defaulted to `numpy==1.11.1` where Python 3.7 requires `numpy==1.14.5`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,8 @@ env:
         - "PS4='+(${BASH_SOURCE}:${LINENO}): ${FUNCNAME[0]:+${FUNCNAME[0]}(): }'"
         # pip dependencies to _test_ your project
         - TEST_DEPENDS="numpy==1.11.1"
+        # params to bdist_wheel. used to set osx build target.
+        - BDIST_PARAMS=""
 
         - PLAT=x86_64
         - UNICODE_WIDTH=32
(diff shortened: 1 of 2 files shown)


skvark:
Nice! Thanks a lot. I suspected that there was something overriding those params. Let's see how the builds go as I have been poking with master branch today.

jcfr:
Thanks @yonip for reporting the issue on scikit-build issue tracker and also thanks for your help to address the issue.

The problem should be fixed in the upcoming release of scikit-build, then specifying either `plat-name macosx-10.7-x86_64` or `-DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.7` should have the expected effect.

Issue #142, opened 2018-11-10
OSX FFmpeg can no longer be built without GPL libs


native-api:
Since https://github.com/Homebrew/homebrew-core/commit/c1057cc939bd2988230af44be4d06b650174ec4e#diff-8f4deb87ce96b9c5efe97be3288bb406 , it's no longer possible to build `ffmpeg` in Homebrew without x264 and x265.

This means that an OSX build built after 2018-10-17 would be covered by GPLv2.

This is also the primary reason for the builds timing out: when you `brew install ffmpeg`, `cmake`, `x265`, `ffmpeg` and a few lesser libraries are built which combined takes a lot more than 20 minutes (in particular, `cmake` takes ~15min, `x265` -- ~20min, `ffmpeg` itself -- ~10min).

Possible solutions are:

* Ask the upstream to reintroduce `--without-gpl`.

    Not sure if a Brew formula can be tuned to not require `x264` and `x265` in this case.

* Build `ffmpeg` from source directly like in Linux.
    * Will need to calculate the dependencies and compiler flags by hand (paths to Brew-installed libs in particular).
    * Also need to detect and install upstream updates by hand.
    * Caching the result is a problem since it will be spread spread around `/usr/local`. [`make dist`](https://www.gnu.org/prep/standards/html_node/Standard-Targets.html) doesn't seem to be available in `ffmpeg`'s `Makefile`.

* Fork the older formula and use it instead of the stock one.
   * Will have to maintain it yourself, of course.
        * Dependencies in particular are a time bomb.
        * Need to detect and install upstream updates by hand.

PR #144, opened 2018-11-15
Cache locally-built Homebrew bottles in OSX builds

native-api:
Closes https://github.com/skvark/opencv-python/issues/139.

* This PR doesn't fix https://github.com/skvark/opencv-python/issues/142 , only addresses the immediate problem of failing builds.

It saves brew dependencies as they are built, measures elapsed time, and once it's high enough, fails the build but arranges for Travis cache to be saved. Then reuses the bottles on the next run. [It doesn't seem that any solution currently exists](https://stackoverflow.com/questions/39930171/cache-brew-builds-with-travis-ci?noredirect=1&lq=1), so rolled my own. Now, it does!
It also correctly detects updates and rebuilds bottles for new versions as needed.

So, basically, you need to restart jobs that print a message like "Building dependencies took too long" in Travis UI until they finish. As things are now, it takes 2 restarts to get a job to finish for the first time.
I'm not sure if Travis UI allows to restart jobs in a PR's automatic check -- in this PR, you'll be able to test this!
* Travis provides [build stages](https://docs.travis-ci.com/user/build-stages/) to avoid the need for restarts, but with 2 restarts, the mac part of the build matrix would grow 3x. Since fixing https://github.com/skvark/opencv-python/issues/142 will reduce build times drastically, I'm reluctant to make such changes until then.
[
Travis looks for cache in the order: PR -> branch -> default branch](https://docs.travis-ci.com/user/caching#fetching-and-storing-caches). So building on `master` should make any other branch be able to find a cache and hopefully finish on the first try.

---

<s>In other news:</s> Other changes:

* Cached the Homebrew installation (but not the cellar) as well so it's not updated all the way each time.
* Set up `ccache` cache. This brings recompile times down from 20 to 8(!) minutes. Dunno how much it will help in the normal course of action 'cuz the effect depends on how many source files are unchanged between builds.
    * [The default `ccache` cache size limit in 1GB](https://linux.die.net/man/1/ccache). Thus it may cause cache uploading timeouts in perspective. I dunno how much data [Travis' 3 min caching timeout](https://docs.travis-ci.com/user/caching#setting-the-timeout) translates into, so can't adjust limits myself upfront.
* Added `-j<...>` to `make` args since [all Travis builders are 2-core](https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system). This only affects the project itself (`brew` uses its own `make` args which already include `-j<...>`, see `brew --env` output).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -524,6 +524,21 @@ before_install: |
     source multibuild_customize.sh
     echo $ENABLE_CONTRIB > contrib.enabled
     echo $ENABLE_HEADLESS > headless.enabled
+    
+    if [ -n "$IS_OSX" ]; then
+        TAPS="$(brew --repository)/Library/Taps"
+        if [ -e "$TAPS/caskroom/homebrew-cask" -a -e "$TAPS/homebrew/homebrew-cask" ]; then
+            rm -rf "$TAPS/caskroom/homebrew-cask"
+        fi
+        find "$TAPS" -type d -name .git -exec \
+                bash -xec '
+                    cd $(dirname '\''{}'\'')
+                    git clean -fxd
+                    git status' \;
+                    
+        brew_cache_cleanup
+    fi
+    
     before_install
     # Not interested in travis internal scripts' output
     set +x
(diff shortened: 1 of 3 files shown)


skvark:
Nice! I will check this during weekend, I'm too busy today/tomorrow.

skvark:
I will check this and the other pull requests / issues asap during this week, didn't found the time last weekend.

skvark:
About that deployment target param: it was a hack due to a bug in scikit-build (https://github.com/skvark/opencv-python/pull/110#issuecomment-408645091). I think it can be moved safely to ``setup.py`` now.

skvark:
I'm a bit afraid of the maintenance burden this might introduce, but I think there's currently no better way to do this. And the the build times are so much better with this approach.

native-api:
> This PR doesn't fix #142

Shucks, Github detected the "fix" word and closed https://github.com/skvark/opencv-python/issues/142 when you merged this, too!
Fortunately, it doesn't matter in this case :)

skvark:
Github is trying to be too smart 😄

PR #145, opened 2018-11-15
remove secret variables

native-api:
Would you consider moving secret variables into Travis settings?

Without secret variables in play, build output is not piped to a filter script, which:

* [gives no buffering, fancy colors and pseudographics in output](https://travis-ci.org/native-api/opencv-python/builds/454101953);
* avoids output loss due to https://github.com/travis-ci/travis-ci/issues/6018. I was hit by it numerous times (see references after https://github.com/travis-ci/travis-ci/issues/6018#issuecomment-402958951);
* `brew` prints some status messages only if `stdout` is a tty, so there's no way to get them with secret variables present.


skvark:
I moved them to Travis settings.

PR #146, opened 2018-11-20
Use patched ffmpeg formula in OSX

native-api:
Closes https://github.com/skvark/opencv-python/issues/142

Includes https://github.com/skvark/opencv-python/pull/144 and is supposed to be applied on top of it.

From options in https://github.com/skvark/opencv-python/issues/142 , I used the "fork the formula" one but not used an older revision but am rather editing the current one. So this can break in perspective if they change the formula file in ways I didn't anticipate.


skvark:
Let's try this. If it ends up breaking too often, we can try to figure some other solution later.

PR #149, opened 2018-11-21
Use ccache

native-api:
Includes https://github.com/skvark/opencv-python/pull/146 and is supposed to be applied on top of it.

Includes a workaround for https://github.com/skvark/opencv-python/issues/148 (which won't be needed once it's solved).

Ccache brings rebuild time from 15 to 5 minutes (for a regular build).

The ccache stats printed at the end are aggregate stats -- since the creation of the cache. I wasn't sure if zeroing them at the start would be more useful.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -524,6 +524,21 @@ before_install: |
     source multibuild_customize.sh
     echo $ENABLE_CONTRIB > contrib.enabled
     echo $ENABLE_HEADLESS > headless.enabled
+    
+    if [ -n "$IS_OSX" ]; then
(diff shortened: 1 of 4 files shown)


skvark:
I think the aggregate stats are ok.

native-api:
Great! [Travis should be now trying to make exactly 3 builds in `master`](https://travis-ci.org/skvark/opencv-python/builds), so after they finish, we should have caches for everything. Even manual restarts won't be needed if there are no irrelevant errors.

skvark:
Thanks for all the hard work again!

PR #150, opened 2018-11-25
Clear yum metadata in x32 docker

native-api:
Closes https://github.com/skvark/opencv-python/issues/148.

I've tried rebuilding the Docker image, and the problem persists. So `yum clean` it is.

The problem doesn't happen (as of now) in the x64 docker image. Maybe it has something to do with x32 emulation, who knows.

**After merging this and https://github.com/skvark/opencv-python/pull/149, revert https://github.com/native-api/opencv-python/commit/8aaff83a74175efbd697c287035350295d6e5b92`.** (I cannot do that here because the reverse commit is only applicable to https://github.com/native-api/opencv-python/tree/use_ccache.)


skvark:
I'll do the revert. I will also rebuild the Docker images at some point and push them to Quay.

PR #182, opened 2019-03-23
Fix OSX builds and migrate to Xcode 8.3

native-api:
Subj. Major changes to react to are:

* In OSX 10.12, Homebrew has a new kind of dependency tracking -- "runtime dependency". E.g. [upgrading `openssl` would now lead to upgrading more than 60 packages linked against the old version](https://discourse.brew.sh/t/can-i-install-a-new-version-without-having-to-upgrade-runtime-dependents/4443).
    * Using [a different bottle install technique](https://discourse.brew.sh/t/can-i-install-a-new-version-without-having-to-upgrade-runtime-dependents/4443/9?u=native-api) now to install new versions alongside the old ones to avoid that.
* Python is now one of `ffmpeg`'s dependencies (indirectly) and its update conflicts with multibuild-installed python. Moreover, Homebrew upgrades the installed Py2 to Py3 as the "default" python.
    * Moved multibuild python installation to after brew build to compensate.
* Newer FFmpeg `brew` formula has more GPLed components.

Other changes are:

* Using Travis build stages to build OSX brew bottles. Now manual job restarts are not needed.

(I also see that you never regenerated x86 Docker images as promised. So Linux x86 builds still run without ccache.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -586,9 +586,10 @@ before_install: |
         fi
         find "$TAPS" -type d -name .git -exec \
                 bash -xec '
-                    cd $(dirname '\''{}'\'')
-                    git clean -fxd
-                    git status' \;
+                    cd $(dirname '\''{}'\'') || echo "status: $?"
+                    git clean -fxd || echo "status: $?"
+                    sleep 1 || echo "status: $?"
+                    git status || echo "status: $?"' \; || echo "status: $?"
 
         brew_cache_cleanup
     fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,6 @@ matrix:
 
     # default builds for MacOS
     - os: osx
-      language: generic
       osx_image: xcode8
       env:
         - MB_PYTHON_VERSION=2.7
@@ -48,7 +47,6 @@ matrix:
         - ENABLE_HEADLESS=0
         - BDIST_PARAMS="-- -DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.7"
     - os: osx
-      language: generic
       osx_image: xcode8
       env:
         - MB_PYTHON_VERSION=3.4
@@ -56,7 +54,6 @@ matrix:
         - ENABLE_HEADLESS=0
         - BDIST_PARAMS="-- -DCMAKE_OSX_DEPLOYMENT_TARGET:STRING=10.7"
     - os: osx
-      language: generic
       osx_image: xcode8
       env:
         - MB_PYTHON_VERSION=3.5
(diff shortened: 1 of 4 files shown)


skvark:
Thank you. I will renegerate the Docker images before merging this, I completely forgot that earlier...

native-api:
Contrib builds fail due to https://github.com/opencv/opencv_contrib/issues/920 . I added a fix there -- you can temporarily switch the submodule to my branch if you don't want to wait for upstream.

skvark:
About that contrib build issue: is Tesseract now a dependency in contrib builds or did the builds fail even without Tesseract? If we disable Tesseract via cmake flags, would the build pass with older opencv_contrib version 4.0.0?

I rebuilt the Docker images yesterday so ccache should work in future builds correctly.

native-api:
Tesseract has been a dependency of contrib module `text` since 3.0.0: https://github.com/opencv/opencv_contrib/commit/5c89c786f2129a0f49f989e28f1f7d99d4201c3d . AFAICS the module _can_ be compiled without it, [OCR functions will print an error message then](https://github.com/opencv/opencv_contrib/search?q=HAVE_TESSERACT&type=Code).

This error only surfaced now because [in the `xcode 8` Travis image, tesseract was not installed judging by the build log of the previous release](https://travis-ci.org/skvark/opencv-python/jobs/472951125#L4181).

I don't see how to explicitly disable tesseract in https://github.com/opencv/opencv_contrib/blob/master/modules/text/CMakeLists.txt , so you'll have to uninstall it, I guess, if you want to build without it.

I don't really see a problem with buiding against the fixed submodule commit because it's just after the release, with no functional differences.

native-api:
> I don't really see a problem with buiding against the fixed submodule commit because it's just after the release, with no functional differences.

I stay corrected, there are, bt only a few: https://github.com/opencv/opencv_contrib/compare/4.0.1...fdb575d43b0f75b8546da313b93892033fc487be

skvark:
Thanks for the information. It's fine to keep it as a dependency since explicitly uninstalling it just adds more work.

And thanks again for this PR. I have been lately too occupied with other stuff and it would have taken me a lot of time to find the root cause to this issue.

skvark:
> I stay corrected, there are, but only a few: opencv/opencv_contrib@4.0.1...fdb575d

It's fine, I'll just add a note to the release that the contrib version does not match exact tag due to the bug.

native-api:
> This error only surfaced now because [in the `xcode 8` Travis image, tesseract was not installed judging by the build log of the previous release](https://travis-ci.org/skvark/opencv-python/jobs/472951125#L4181).

On a closer inspection, it's a dependency of the `ffmpeg` formula since https://github.com/Homebrew/homebrew-core/commit/3e99e5d2717c30ebc1f555d9d70b08cf6b6d33f8

native-api:
> On a closer inspection, it's a dependency of the `ffmpeg` formula since [Homebrew/homebrew-core@3e99e5d](https://github.com/Homebrew/homebrew-core/commit/3e99e5d2717c30ebc1f555d9d70b08cf6b6d33f8)

For the record, this means that it can be excluded when generating the custom formula, the same way as GPL libraries.

PR #227, opened 2019-07-30
Travis CI: Xenial is now the default Ubuntu in Travis

cclauss:
https://blog.travis-ci.com/2019-04-15-xenial-default-build-environment

Also: __sudo: required__ no longer is [Travis are now recommending removing the __sudo__ tag](https://blog.travis-ci.com/2018-11-19-required-linux-infrastructure-migration) because the sudo command is always available in Travis CI and there is no longer any way to turn it off.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,11 +13,9 @@ env:
 
 # Host Python is never used
 language: generic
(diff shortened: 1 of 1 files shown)


skvark:
Thanks!

PR #228, opened 2019-07-30
Use Travis workspaces

native-api:
[A new Travis CI feature, workspaces](https://docs.travis-ci.com/user/using-workspaces/), allows to use one MacOS cache job to build FFMpeg instead of 20.

It's currently in beta -- so it's your call whether it's good enough for production use. [It worked for me well](https://travis-ci.org/native-api/opencv-python/builds/560299543) save for a [small nitpick](https://travis-ci.community/t/introducing-workspaces/4249/3?u=native-api) when I used this project to try it out.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -239,8 +239,6 @@ matrix:
 
     # default builds for Linux
     - os: linux
-      # the following jobs will use the same stage name by default
-      stage: s1
       env:
         - MB_PYTHON_VERSION=2.7
         - ENABLE_CONTRIB=0
--- a/.travis.yml
+++ b/.travis.yml
@@ -52,6 +52,17 @@ matrix:
             - /usr/local/Homebrew/
             # used in OSX custom build script dealing with local bottle caching
             - $HOME/local_bottle_metadata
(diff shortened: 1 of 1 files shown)


native-api:
It shaves about 11 hours off of full build time (~5,5 hours of real time since 2 OSX jobs run in parallel).

native-api:
On a closer inspection, it definitely shaves off machine time massively, but doesn't save real time much because the Linux jobs are now the critical path, and full-building the 20 OSX jobs takes about the same time as the 48 Linux ones.

Moving them to the Final stage would fix this, but running Linux builds will be delayed by ~40 min (~10 min for rebuilds) that the S1 stage takes -- and they will not run at all if it fails.

Is this acceptable? If you need to debug some issues with a specific job, you can disable everything else temporarily anyway (and then edit the history to remove those temporary commits).

skvark:
I think it's ok. In any case it's usually best to disable all but one or two jobs if there are issues. Also the Linux side is far more easier to debug than macOS because it can be done locally via Docker.

skvark:
Let's give this a try.

Issue #193, opened 2019-04-24
Python: YouTube livestream not working with Streamlink


devanshusharma19:
### Expected behaviour
opencv working with youtube stream or youtube videos

### Actual behaviour
actually it is not working it is showing error. I have tried on browser that link. It is working fine but in opencv is not working.

> cv2.error: OpenCV(4.0.1) /io/opencv/modules/highgui/src/window.cpp:352: error: (-215:Assertion failed) size.width>0 && size.height>0 in function 'imshow' 

### Steps to reproduce

- example code
```
import streamlink
import cv2

url = 'https://www.youtube.com/watch?v=bZ6NL59FMoc'

streams = streamlink.streams(url)
print(streams)
#cap = cv2.VideoCapture(streams["360p"].url)
cap = cv2.VideoCapture(streams["best"].url)

while True:
    ret, frame = cap.read()
    print(frame)
    cv2.imshow('frame', frame)

    if cv2.waitKey(1) & 0xff == ord('q'):
        break

cap.release()
cv2.destroyAllWindows()
```
- operating system

> Ubuntu 18.04

- architecture (e.g. x86)

> 64 bit
>python 3.6

- opencv-python version

> opencv-python version 4.0.1

Sushanti99:
Hello @devanshusharma19 
This code works perfectly for me on Windows. 
You can try manually fixing the height and width of the frame on Ubuntu I guess.

native-api:
Looks like a duplicate of https://github.com/skvark/opencv-python/issues/204

fspider:
Hello
With some youtube videos,this code works.
But this code not works with youtube live stream.
(https://www.youtube.com/watch?v=17Deeq8N2e4)

1. Method
        vPafy = pafy.new(videoUrl)
        play = vPafy.getbest(preftype="mp4")
        return play.url

2. Method
        streams = streamlink.streams(videoUrl)
        return streams["best"].url

All 2 Method not works.

The error screen like this

[ERROR:0] global C:\projects\opencv-python\opencv\modules\videoio\src\cap.cpp (116) cv::VideoCapture::open VIDEOIO(CV_IMAGES): raised OpenCV exception:

[...]

PR #229, opened 2019-07-30
Added OpenSSL & various protocol support to FFmpeg backend

### LINKED ISSUE #204: Bug: https URL not working with opencv-python
## Bug summary:
`opencv-contrib-python` python library's `VideoCapture` class  fails to parse https URL received from `pafy/youtube-dl`. But strangely this bug doesn't exist on same/different machine with OpenCV installed from scratch.

### Affected users:
Users with OpenCV installed with [`opencv-contrib-python`](https://pypi.org/project/opencv-contrib-python/) or [`opencv-python`](https://pypi.org/project/opencv-python/) python library from PyPi

### Non-Affected users:
Users with [OpenCV installed from scratch/source-code](https://www.pyimagesearch.com/2018/05/28/ubuntu-18-04-how-to-install-opencv/) and not from PyPi (Tested with OpenCV 4.1.0-dev & 3.4.5)

### Steps to reproduce
- operating system - `Ubuntu 18.04.1 bionic `
- architecture `x64`
- opencv-python           ` 4.1.0.25` 
- youtube-dl               `2019.5.20`  
- pafy                    ` 0.5.4`    
- python_version  `3.6/2.7/3.5 tested`

### Test-Code: 

**You can copy `url` directly to `cv2.VideoCapture` class and this result in same Bug.**

```python
import cv2
import pafy

video = pafy.new("dQw4w9WgXcQ")
object = video.getbestvideo()
url = object.url
print(url) #means pafy and youtube-dl working correctly

stream = cv2.VideoCapture(url)

# infinite loop
while True:
	(grabbed, frame) = stream.read()
	# read frames

	# check if frame empty
	if not grabbed:
		print('opencv is buggy')
		break

	cv2.imshow("Output Frame", frame)
	# Show output window

	key = cv2.waitKey(1) & 0xFF
	# check for 'q' key-press
	if key == ord("q"):
		#if 'q' key-pressed break out
		break

cv2.destroyAllWindows()
# close output window

stream.release()
# safely close video stream.
``` 

**This code exit instantly with no error on `opencv-python 4.1.0.25` since no frame grabbed BUT instead this algorithm works when OpenCV installed from scratch and returns valid frames.**

abhiTronix:
## PR Description
This PR introduces Major Fixes and updates like OpenSSL for opencv-python's FFmpeg backend.

### Fixes
#204 
#193 

### Changes
- Major Fixes and updates for Docker files (both x86 & x86-64)
- Enabled FFmpeg backend with OpenSSL
- Removed `libjpeg-turbo` libs build from scratch

### DIFF
--- a/docker/Dockerfile_i686
+++ b/docker/Dockerfile_i686
@@ -1,5 +1,7 @@
 FROM quay.io/pypa/manylinux1_i686:latest
 
+ENTRYPOINT ["linux32", "--"]
+
 RUN curl -O -L https://download.qt.io/archive/qt/4.8/4.8.7/qt-everywhere-opensource-src-4.8.7.tar.gz && \
 	tar -xf qt-everywhere-opensource-src-4.8.7.tar.gz && \
 	cd qt-everywhere* && \
--- a/docker/Dockerfile_i686
+++ b/docker/Dockerfile_i686
@@ -1,7 +1,5 @@
 FROM quay.io/pypa/manylinux1_i686:latest
 
-ENTRYPOINT ["linux32", "--"]
-
 RUN curl -O -L https://download.qt.io/archive/qt/4.8/4.8.7/qt-everywhere-opensource-src-4.8.7.tar.gz && \
 	tar -xf qt-everywhere-opensource-src-4.8.7.tar.gz && \
 	cd qt-everywhere* && \
@@ -16,56 +14,71 @@ ENV QTDIR /opt/Qt4.8.7
 ENV PATH "$QTDIR/bin:$PATH"
 
 RUN curl -O -L https://cmake.org/files/v3.9/cmake-3.9.0.tar.gz && \
-	tar -zxf cmake-3.9.0.tar.gz && \
+	tar -xf cmake-3.9.0.tar.gz && \
 	cd cmake-3.9.0 && \
-	yum -y install curl-devel zlib-devel && \
+    true '#manylinux1 provides curl-devel equivalent and libcurl statically linked \
+         against the same newer OpenSSL as other source-built tools \
+         (1.0.2s as of this writing)' && \
+	yum -y install zlib-devel && \
 	./configure --system-curl && \
 	make -j4 && \
 	make install && \
 	cd .. && \
 	rm -rf cmake-3.9.0*
 
-RUN	yum install autoconf automake bzip2 git freetype-devel gcc gcc-c++ libtool make pkgconfig -y && \
-    curl -O -L https://www.cpan.org/src/5.0/perl-5.20.1.tar.gz && \
-	tar -xzf perl-5.20.1.tar.gz && \
+# https://trac.ffmpeg.org/wiki/CompilationGuide/Centos#GettheDependencies
+# manylinux provides the toolchain and git; we provide cmake
+RUN yum install freetype-devel bzip2-devel zlib-devel -y && \
+    mkdir ~/ffmpeg_sources
+
+# Newer openssl configure requires newer perl
+RUN curl -O -L https://www.cpan.org/src/5.0/perl-5.20.1.tar.gz && \
+	tar -xf perl-5.20.1.tar.gz && \
 	cd perl-5.20.1 && \
-	./Configure -des -Dprefix="$HOME/ffmpeg_build" && \
-	make -j4 && \
-	make install && \
+	./Configure -des -Dprefix="$HOME/openssl_build" && \
+	true '#perl build scripts do much redundant work \
+	     if running "make install" separately' && \
+	make install -j4 && \
 	cd .. && \
 	rm -rf perl-5.20.1*
 
-RUN	yum remove nasm -y && \
-	mkdir ~/ffmpeg_sources && \
-	cd ~/ffmpeg_sources && \
+RUN cd ~/ffmpeg_sources && \
+	curl -O -L https://github.com/openssl/openssl/archive/OpenSSL_1_1_1c.tar.gz && \
+	tar -xf OpenSSL_1_1_1c.tar.gz && \
+	cd openssl-OpenSSL_1_1_1c && \
+    true '#in i686, ./config detects x64 in i686 without linux32 \
+         when run from "docker build"' && \
+	PERL="$HOME/openssl_build/bin/perl" linux32 ./config --prefix="$HOME/ffmpeg_build" --openssldir="$HOME/ffmpeg_build" shared zlib && \
+	make -j4 && \
+    true '#skip installing documentation' && \
+	make install_sw && \
+    rm -rf ~/openssl_build
+
+RUN cd ~/ffmpeg_sources && \
 	curl -O -L http://www.nasm.us/pub/nasm/releasebuilds/2.13.02/nasm-2.13.02.tar.bz2 && \
-	tar xjvf nasm-2.13.02.tar.bz2 && cd nasm-2.13.02 && ./autogen.sh && \
+	tar -xf nasm-2.13.02.tar.bz2 && cd nasm-2.13.02 && ./autogen.sh && \
 	./configure --prefix="$HOME/ffmpeg_build" --bindir="$HOME/bin" && \
 	make -j4 && \
-	make install && \
-	cd ~/ffmpeg_sources && \
+	make install
+
+RUN cd ~/ffmpeg_sources && \
 	curl -O -L http://www.tortall.net/projects/yasm/releases/yasm-1.3.0.tar.gz && \
-	tar xzvf yasm-1.3.0.tar.gz && \
+	tar -xf yasm-1.3.0.tar.gz && \
 	cd yasm-1.3.0 && \
 	./configure --prefix="$HOME/ffmpeg_build" --bindir="$HOME/bin" && \
 	make -j4 && \
-	make install && \
-	cd ~/ffmpeg_sources && \
+	make install
+
+RUN cd ~/ffmpeg_sources && \
 	git clone --depth 1 https://chromium.googlesource.com/webm/libvpx.git && \
 	cd libvpx && \
 	./configure --prefix="$HOME/ffmpeg_build" --disable-examples --disable-unit-tests --enable-vp9-highbitdepth --as=yasm --enable-pic --enable-shared && \
 	make -j4 && \
-	make install && \
-	cd ~/ffmpeg_sources && \
-	curl -O -L https://github.com/openssl/openssl/archive/OpenSSL_1_1_1c.tar.gz && \
-	tar -zxvf OpenSSL_1_1_1c.tar.gz && \
-	cd openssl-OpenSSL_1_1_1c && \
-	PERL="$HOME/ffmpeg_build/bin/perl" ./config --prefix="$HOME/ffmpeg_build" --openssldir="$HOME/ffmpeg_build" shared zlib && \
-	make -j4 && \
-	make install_sw && \
-	cd ~/ffmpeg_sources && \
+	make install
+
+RUN cd ~/ffmpeg_sources && \
 	curl -O -L https://ffmpeg.org/releases/ffmpeg-snapshot.tar.bz2 && \
-	tar xjvf ffmpeg-snapshot.tar.bz2 && \
+	tar -xf ffmpeg-snapshot.tar.bz2 && \
 	cd ffmpeg && \
 	PATH=~/bin:$PATH && \
 	PKG_CONFIG_PATH="$HOME/ffmpeg_build/lib/pkgconfig" ./configure --prefix="$HOME/ffmpeg_build" --extra-cflags="-I$HOME/ffmpeg_build/include" --extra-ldflags="-L$HOME/ffmpeg_build/lib" --enable-openssl --enable-libvpx --enable-shared --enable-pic --bindir="$HOME/bin" && \
@@ -78,30 +91,15 @@ RUN	yum remove nasm -y && \
 ENV PKG_CONFIG_PATH /usr/local/lib/pkgconfig:/root/ffmpeg_build/lib/pkgconfig
 ENV LDFLAGS -L/root/ffmpeg_build/lib
 
-RUN mkdir libjpeg-turbo && \
-	cd libjpeg-turbo && \
-	export PATH=~/bin:$PATH && \
-	curl -L https://kent.dl.sourceforge.net/project/libjpeg-turbo/1.5.3/libjpeg-turbo-1.5.3.tar.gz > libjpeg-turbo-1.5.3.tar.gz && \
-	tar xzvf libjpeg-turbo-1.5.3.tar.gz && \
-	cd libjpeg-turbo-1.5.3 && \
-	export CFLAGS="-fPIC" && \
-	export CXXFLAGS="-fPIC" && \
-	autoreconf -fiv && \
-	./configure --host=i686-pc-linux-gnu && \
-	make && \
-	make install && \
-	cd ../../ && \
-	rm -rf libjpeg-turbo
-
-ENV JPEG_LIBRARY /opt/libjpeg-turbo/lib32/libjpeg.a
-ENV JPEG_INCLUDE_DIR /opt/libjpeg-turbo/include
-
 RUN curl -O https://raw.githubusercontent.com/torvalds/linux/v4.14/include/uapi/linux/videodev2.h && \
 	curl -O https://raw.githubusercontent.com/torvalds/linux/v4.14/include/uapi/linux/v4l2-common.h && \
 	curl -O https://raw.githubusercontent.com/torvalds/linux/v4.14/include/uapi/linux/v4l2-controls.h && \
 	curl -O https://raw.githubusercontent.com/torvalds/linux/v4.14/include/linux/compiler.h && \
 	mv videodev2.h v4l2-common.h v4l2-controls.h compiler.h /usr/include/linux
 
+#in i686, yum metadata ends up with slightly wrong timestamps
+#which inhibits its update
+#https://github.com/skvark/opencv-python/issues/148
 RUN yum clean all
 
 ENV PATH "$HOME/bin:$PATH"
--- a/docker/Dockerfile_i686
+++ b/docker/Dockerfile_i686
@@ -45,11 +45,11 @@ RUN yum install autoconf automake bzip2 cmake freetype-devel gcc gcc-c++ libtool
 	./configure --prefix="$HOME/ffmpeg_build" --disable-examples --disable-unit-tests --enable-vp9-highbitdepth --as=yasm --enable-pic --enable-shared && \
 	make -j4 && \
 	make install && \
-        cd ~/ffmpeg_sources && \
+    cd ~/ffmpeg_sources && \
 	curl -O -L https://github.com/openssl/openssl/archive/OpenSSL_1_1_1c.tar.gz && \
 	tar -zxvf OpenSSL_1_1_1c.tar.gz && \
 	cd openssl-OpenSSL_1_1_1c && \
-	./configure --prefix="$HOME/ffmpeg_build" --openssldir="$HOME/ffmpeg_build" shared zlib && \
+	./Configure --prefix="$HOME/ffmpeg_build" --openssldir="$HOME/ffmpeg_build" shared zlib && \
 	make -j4 && \
 	make install && \
 	cd ~/ffmpeg_sources && \
--- a/docker/Dockerfile_i686
+++ b/docker/Dockerfile_i686
@@ -13,18 +13,28 @@ RUN curl -O -L https://download.qt.io/archive/qt/4.8/4.8.7/qt-everywhere-opensou
 ENV QTDIR /opt/Qt4.8.7
 ENV PATH "$QTDIR/bin:$PATH"
 
-RUN curl -O -L https://cmake.org/files/v3.9/cmake-3.9.0.tar.gz && \
-	tar -zxf cmake-3.9.0.tar.gz && \
-	cd cmake-3.9.0 && \
-	yum -y install curl-devel zlib-devel && \
-	./configure --system-curl && \
-	make && \
-	make install && \
-	cd .. && \
+RUN curl -O -L https://cmake.org/files/v3.9/cmake-3.9.0.tar.gz  && \
+	tar -zxf cmake-3.9.0.tar.gz  && \
+	cd cmake-3.9.0  && \
+	yum -y install curl-devel zlib-devel  && \
+	./configure --system-curl  && \
+	make -j4 && \
+	make install  && \
+	cd ..  && \
 	rm -rf cmake-3.9.0*
 
-RUN yum install autoconf automake bzip2 cmake freetype-devel gcc gcc-c++ libtool make mercurial pkgconfig perl-core zlib-devel -y && \
-	yum remove nasm -y && \
+RUN curl -O -L  https://www.cpan.org/src/5.0/perl-5.10.1.tar.gz  && \
+	tar -xzf perl-5.10.1.tar.gz && \
+	yum install autoconf automake bzip2 git freetype-devel gcc gcc-c++ libtool make pkgconfig  -y  && \
+	cd perl-5.10.1  && \
+	./Configure -des -Dprefix=$HOME/ffmpeg_build && \
+	make -j4 && \
+	make install && \
+	cd ..  && \
+	rm -rf perl-5.10.1* && \
+	PATH=$HOME/ffmpeg_build/bin:$PATH #otherwise perl not detected at runtime
+
+RUN	yum remove nasm -y && \
 	mkdir ~/ffmpeg_sources && \
 	cd ~/ffmpeg_sources && \
 	curl -O -L http://www.nasm.us/pub/nasm/releasebuilds/2.13.02/nasm-2.13.02.tar.bz2 && \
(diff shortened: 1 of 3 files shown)


abhiTronix:
@native-api I tested these changes on my machine and this fixes my [bug](https://github.com/skvark/opencv-python/issues/204#issue-446912322) too. Also, I didn't use `--enable-non-free` to avoid license issues and It seems to works without it. :+1:

skvark:
Looks good. I have to build the Docker images locally and push them to quay.io. I'll try to make new releases during this month.

docker/Dockerfile_i686
```diff
@@ -23,7 +23,7 @@ RUN curl -O -L https://cmake.org/files/v3.9/cmake-3.9.0.tar.gz && \
 	cd .. && \
 	rm -rf cmake-3.9.0*
 
-RUN yum install autoconf automake bzip2 cmake freetype-devel gcc gcc-c++ libtool make mercurial pkgconfig zlib-devel -y && \
+RUN yum install autoconf automake bzip2 cmake freetype-devel gcc gcc-c++ libtool make mercurial pkgconfig openssl-devel zlib-devel -y && \
```
native-api: CentOS 5's `openssl` is 0.9.8e which is ancient. Is it safe to use?..

docker/Dockerfile_i686
```diff
@@ -50,7 +50,7 @@ RUN yum install autoconf automake bzip2 cmake freetype-devel gcc gcc-c++ libtool
 	tar xjvf ffmpeg-snapshot.tar.bz2 && \
 	cd ffmpeg && \
 	PATH=~/bin:$PATH && \
-	PKG_CONFIG_PATH="$HOME/ffmpeg_build/lib/pkgconfig" ./configure --prefix="$HOME/ffmpeg_build" --extra-cflags="-I$HOME/ffmpeg_build/include" --extra-ldflags="-L$HOME/ffmpeg_build/lib" --enable-libvpx --enable-shared --enable-pic --bindir="$HOME/bin" && \
+	PKG_CONFIG_PATH="$HOME/ffmpeg_build/lib/pkgconfig" ./configure --prefix="$HOME/ffmpeg_build" --extra-cflags="-I$HOME/ffmpeg_build/include" --extra-ldflags="-L$HOME/ffmpeg_build/lib" --enable-libvpx --enable-openssl --enable-protocol=file,ftp,http,https,httpproxy,hls,mmsh,mmst,pipe,rtmp,rtmps,rtmpt,rtmpts,rtp,sctp,srtp,tcp,udp --enable-shared --enable-pic --bindir="$HOME/bin" && \
```
native-api: Is there a particular reason to enable these specific protocols? By default, all supported protocols are enabled AFAICS.

native-api: (Not criticizing; just wondering since this looks like just a copied full list but it's not actually the full list.)

abhiTronix: > Is there a particular reason to enable these specific protocols? By default, all supported protocols are enabled AFAICS.

@native-api Thanks for pointing this out. No there is no specific reason, I just thought these are the most commonly used protocols. Should I remove them?

docker/Dockerfile_i686
abhiTronix: > CentOS 5's openssl is 0.9.8e which is ancient. Is it safe to use?..

@native-api Wow, I didn't know that will be a problem too. Anyhow, I found [this](https://blacksaildivision.com/how-to-install-openssl-on-centos) simple tutorial to do that. Should I make these changes?

docker/Dockerfile_i686
native-api: Yes. I don't think we should change the default without a real reason (I'm not implying that there's none, just that none was presented).

docker/Dockerfile_i686
native-api: I'd go the same way other ffmpeg dependencies are built but that guide looks like a good reference point. Can't say much more, it's your job to see that everything hooks up nicely.

abhiTronix: @native-api Does this library links system FFmpeg in any case or only use this precompiled FFmpeg? I'm not sure if this patch worked on my machine due to these changes or it used my system FFmpeg during compiling while testing.

abhiTronix:
@skvark Kindly review new changes/commit I recently added. Thank you.

native-api:
@abhiTronix Did you test your changes? The x64 dockerfile at least fails for me with `/configure: command not found` because the script in https://github.com/openssl/openssl/archive/OpenSSL_1_1_1c.tar.gz is called "`config`"; and `perl-core` fails to install because there's no such package.

To make it clear when you are ready to show your changes, you can make a branch for testing and merge it to the PR branch when you are ready.

abhiTronix:
> Did you test your changes?

Yes, there were no errors on my machine(Ubuntu 16.04 x64).

> The x64 dockerfile at least fails for me with ./configure: No such file or directory because the script in https://github.com/openssl/openssl/archive/OpenSSL_1_1_1c.tar.gz is called "config"

But all other commands in docker is called with `./configure` command. We use this configure script  for getting ready to build the software on linux specific system. It makes sure all of the dependencies for the rest of the build and install process are available, and finds out whatever it needs to know to use those dependencies. This is a problem with OpenSSL itself to rename that file, not mistake from my end. Anyways I'll correct it.

>  perl-core fails to install because there's no such package.

On what machine are you testing? this package is meant for CentOS, and it's a [valid package](https://centos.pkgs.org/7/centos-x86_64/perl-core-5.16.3-293.el7.x86_64.rpm.html) FYI.

> To make it clear when you are ready to show your changes, you can make a branch for testing and merge it to the PR branch when you are ready.

Sorry but another PR will be waste of time. If you think something is wrong, Kindly point me in that direction to solve it rather than making it more confusing. Thank you.

native-api:
Commands specified in a Dockerfile do not run directly on your system. They instead run in the container specified in `FROM`. So you need to test them [with `docker build - <Dockerfile`](https://docs.docker.com/engine/reference/commandline/build/#text-files).
For debugging and experimentation, you can `docker run -it <container in FROM> bash --login` and run the commands manually in the resulting shell.

skvark:
The ``manylinux`` base image CentOS version is 5. It has to be very old to make the resulting binaries compatible with as many as possible glibc based GNU/Linux distributions. Therefore many dependencies have to be built manually because the pre-built packages provided by CentOS repositories can be easily 10 years old or they do not simply exist in there.

I originally assumed that the ``manylinux`` image would have contained a fresh build of OpenSSL but apparently it does not. @abhiTronix, please follow @native-api's instructions on how to test the commands. What I usually do:

1. ``docker run -it <container in FROM> bash``
2. Build the new library / dependency manually according to the instructions provided
3. When everything seems to be ok, copy the commands to the Dockerfile in correct order
4. Build the Docker image with ``docker build`` (note: this takes a lot of time even with a fast processor)
5. Run the new container similarly as in step 1. and check that everything is ok

One addition (if possible): I would like to see also the output of ``ffmpeg -L`` command to see that the license is still LGPL.

abhiTronix:
@skvark Thank you for explaining thoroughly.
But, When I tried to run docker commands individually within centos image pulled with `docker run -it centos bash` command, I immediately encountered various path problems and thereby all binaries compiled (like yasm, nasm) were unable to detect while compiling libvpx binary or FFmpeg binaries. I tried various methods to solve it but finally adding `$HOME/ffmpeg_build` and `$HOME/bin` to system path solved this problem. **This is normal behavior or a bug?**

native-api:
> image pulled with `docker run -it centos bash` command

Not `centos` but the image specified in FROM in the dockerfile -- i.e. `quay.io/pypa/manylinux1_x86_64:latest` or `quay.io/pypa/manylinux1_i686:latest`. It has many additional software installed compared to a stock CentOS 5 installation (see https://github.com/pypa/manylinux/tree/manylinux1#readme and https://github.com/pypa/manylinux/tree/manylinux1/docker if you want to know what exactly).

abhiTronix:
@native-api Thanks

I found some inconsistencies in dockers:
1. why reinstalling old cmake in dependencies here:

https://github.com/skvark/opencv-python/blob/da7d02271d2ffca8f293ed3f3032f30881319db8/docker/Dockerfile_x86_64#L26

 when we are building cmake from scratch?

2. `mercurial` is not supported for centos 5 says "No package mercurial available, then why it's a dependency
"
3.  OpenSSL is already a dependency for other vital packages like `curl-devel & zlib-devel` therefore cannot be removed:                               
```shell
Dependencies Resolved

============================================================================================================================================================================================================
 Package                                               Arch                                     Version                                                  Repository                                    Size
============================================================================================================================================================================================================
Installing:
 curl-devel                                            x86_64                                   7.15.5-17.el5_9                                          base                                         318 k
 zlib-devel                                            x86_64                                   1.2.3-7.el5                                              base                                         103 k
Installing for dependencies:
 e2fsprogs-devel                                       x86_64                                   1.39-37.el5                                              base                                         639 k
 keyutils-libs-devel                                   x86_64                                   1.2-1.el5                                                base                                          27 k
 krb5-devel                                            x86_64                                   1.6.1-80.el5_11                                          updates                                      1.9 M
 libidn-devel                                          x86_64                                   0.6.5-1.1                                                base                                         239 k
 libselinux-devel                                      x86_64                                   1.33.4-5.7.el5.centos                                    libselinux                                   149 k
 libsepol-devel                                        x86_64                                   1.15.2-3.el5                                             base                                         192 k
 openssl-devel                                         x86_64                                   0.9.8e-40.el5_11                                         updates                                      1.9 M

Transaction Summary
============================================================================================================================================================================================================
Install       9 Package(s)
Upgrade       0 Package(s)

```

native-api:
`cmake` and `mercurial` from yum look like old leftovers and can safely be removed AFAICS.
`openssl-devel` must be a dependency of `curl-devel`. I don't know if we link anything to `libcurl` -- try removing it and see it something requires it; if something does then we can decide if anything needs to be done about the old `openssl` there. Good find!

native-api:
Oh, I see, it's the built `cmake` that requires `libcurl`.

It's used in the wheel build process, and that does include downloading stuff like IPP IW from OpenCV's supplemental GitHub repos. So the aged `openssl` is not a critical security problem (since it's not present in the wheels) but still a time bomb.

It's very possible however that we'll drop `manylinux1` support before it becomes unable to download it. I see that manylinux1 dockerfiles build git against the same system `openssl`. **So we can rule that `manylinux1` build environment is not designed to last longer that CentOS openssl's, and we don't need to update it if we only use it in build tools.**

abhiTronix:
@native-api Also old system `perl` is not supported by openssl, says 
```shell
Perl v5.10.0 required--this is only v5.8.8, stopped at ./Configure line 12.
BEGIN failed--compilation aborted at ./Configure line 12.
```
We need to build `perl` from scratch too!

native-api:
(Scratch my previous message about a ruling, found new evidence.)
`manylinux1` dockerfile scripts build a newer version of `openssl`. They don't build a newer `libcurl` though.
So for programs that are used solely in the build process like `cmake`, we can link against that version since `manylinux1`'s toolchain is not designed to last beyond its lifespan. You'd probably need to build a `libcurl` against it though.

The toolchain cleanup should better be done in a different PR though. Let's only do what is relevant to the fresh OpenSSL here.

docker/Dockerfile_i686
```diff
@@ -45,12 +45,19 @@ RUN yum install autoconf automake bzip2 cmake freetype-devel gcc gcc-c++ libtool
 	./configure --prefix="$HOME/ffmpeg_build" --disable-examples --disable-unit-tests --enable-vp9-highbitdepth --as=yasm --enable-pic --enable-shared && \
 	make -j4 && \
 	make install && \
+    cd ~/ffmpeg_sources && \
+	curl -O -L https://github.com/openssl/openssl/archive/OpenSSL_1_1_1c.tar.gz && \
+	tar -zxvf OpenSSL_1_1_1c.tar.gz && \
+	cd openssl-OpenSSL_1_1_1c && \
+	./Configure --prefix="$HOME/ffmpeg_build" --openssldir="$HOME/ffmpeg_build" shared zlib && \
```
native-api: OpenSSL's `README` says that `config` is the script that should be run by the user.

abhiTronix: yes, that was a mistake, I changed it while testing.

abhiTronix:
> The toolchain cleanup should better be done in a different PR though. Let's only do what is relevant to the fresh OpenSSL here.

@native-api  but latest OpenSSL build requires at least `Perl v5.10.0` which is not present in CentOS which provide old/obsolete version v5.8.8.

native-api:
> We need to build `perl` from scratch too!
> but latest OpenSSL build requires at least Perl v5.10.0

Use the minimal configuration that works (openssl'd build docs seem to say which Perl modules are needed), we don't need this chain reaction to continue!
Since we don't need the new `perl` for anything but running the script, maybe it's better to use a custom prefix for it.

Perl 5.10 is not a part of the toolchain, it's effectively a private dependency of openssl. You can as well delete it after building the library! So it is relevant. I was talking about other unrelated inconsistencies that you spotted.

abhiTronix:
@skvark License is as follows with the recent build(`ffmpeg -L`): 

```shell
[root@0122eceb268d ffmpeg]# ffmpeg -L
ffmpeg version N-94451-ga3adc3b Copyright (c) 2000-2019 the FFmpeg developers
  built with gcc 4.8.2 (GCC) 20140120 (Red Hat 4.8.2-15)
  configuration: --prefix=/root/ffmpeg_build --extra-cflags=-I/root/ffmpeg_build/include --extra-ldflags=-L/root/ffmpeg_build/lib --enable-openssl --enable-libvpx --enable-shared --enable-pic --bindir=/root/bin
  libavutil      56. 33.100 / 56. 33.100
  libavcodec     58. 55.100 / 58. 55.100
  libavformat    58. 30.100 / 58. 30.100
  libavdevice    58.  9.100 / 58.  9.100
  libavfilter     7. 58.100 /  7. 58.100
  libswscale      5.  6.100 /  5.  6.100
  libswresample   3.  6.100 /  3.  6.100
ffmpeg is free software; you can redistribute it and/or
modify it under the terms of the GNU Lesser General Public
License as published by the Free Software Foundation; either
version 2.1 of the License, or (at your option) any later version.

ffmpeg is distributed in the hope that it will be useful,
but WITHOUT ANY WARRANTY; without even the implied warranty of
MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the GNU
Lesser General Public License for more details.

You should have received a copy of the GNU Lesser General Public
License along with ffmpeg; if not, write to the Free Software
Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA

```

docker/Dockerfile_i686
```diff
@@ -13,18 +13,28 @@ RUN curl -O -L https://download.qt.io/archive/qt/4.8/4.8.7/qt-everywhere-opensou
...
+RUN curl -O -L  https://www.cpan.org/src/5.0/perl-5.10.1.tar.gz  && \
+	tar -xzf perl-5.10.1.tar.gz && \
+	yum install autoconf automake bzip2 git freetype-devel gcc gcc-c++ libtool make pkgconfig  -y  && \
+	cd perl-5.10.1  && \
+	./Configure -des -Dprefix=$HOME/ffmpeg_build && \
+	make -j4 && \
+	make install && \
+	cd ..  && \
+	rm -rf perl-5.10.1* && \
+	PATH=$HOME/ffmpeg_build/bin:$PATH #otherwise perl not detected at runtime
```
native-api: No need to add `perl` to global path for all further commands -- it's possible to override envvars for a single command, see https://github.com/abhiTronix/opencv-python/pull/2

docker/Dockerfile_i686
```diff
@@ -13,18 +13,28 @@ RUN curl -O -L https://download.qt.io/archive/qt/4.8/4.8.7/qt-everywhere-opensou
...
+	make -j4 && \
+	make install  && \
+	cd ..  && \
 	rm -rf cmake-3.9.0*
 
-RUN yum install autoconf automake bzip2 cmake freetype-devel gcc gcc-c++ libtool make mercurial pkgconfig zlib-devel -y && \
-	yum remove nasm -y && \
+RUN curl -O -L  https://www.cpan.org/src/5.0/perl-5.10.1.tar.gz  && \
+	tar -xzf perl-5.10.1.tar.gz && \
+	yum install autoconf automake bzip2 git freetype-devel gcc gcc-c++ libtool make pkgconfig  -y  && \
```
native-api: `i686` Dockerfile fails here due to dependency resolution problems for `git`.

There's no need to install Git because manylinux already provides it.

docker/Dockerfile_i686
```diff
@@ -13,18 +13,28 @@ RUN curl -O -L https://download.qt.io/archive/qt/4.8/4.8.7/qt-everywhere-opensou
...
+	cd perl-5.10.1  && \
+	./Configure -des -Dprefix=$HOME/ffmpeg_build && \
```
native-api: When I fix the `yum install git` problem, OpenSSL build fails in `i686` Dockerfile. Output suggests that Perl has been built for x64 instead of x86.

docker/Dockerfile_i686
abhiTronix: Thanks for this.

docker/Dockerfile_i686
abhiTronix: But builds were failing on my machine due to the absence of `git` on x64 docker? Please wait I will test your PR again.

docker/Dockerfile_i686
abhiTronix: **Please try this for x64:** `./Configure -des -Dccflags=-m64 -Dldflags=-m64 -Dlddlflags="-m64" -Dcc=gcc -Dprefix=$HOME/ffmpeg_build`

**and for x32:** `./Configure -des -Dccflags=-m32 -Dldflags=-m32 -Dlddlflags="-m32" -Dcc=gcc -Dprefix=$HOME/ffmpeg_build`

That way we can force forcing 64-bit mode and 32bit mode

native-api:
I give you carte blanche for including any general dockerfile cleanups into this PR if that makes your job easier.

abhiTronix:
> Operating system: x86_64-whatever-linux2

**@native-api shouldn't it be x86-whatever-linux2 instead ?**

abhiTronix:
>     then OpenSSL compilation in i686 fails again, seemingly again due to wrong architecture:
> 
> Operating system: x86_64-whatever-linux2
> Configuring OpenSSL version 1.1.1c (0x1010103fL) for linux-x86_64
> <...>
> apps/apps.c:1:0: sorry, unimplemented: 64-bit mode not compiled in
>  /*
>  ^
> apps/bf_prefix.c:1:0: sorry, unimplemented: 64-bit mode not compiled in
>  /*
>  ^
> apps/app_rand.c:1:0: sorry, unimplemented: 64-bit mode not compiled in
>  /*
>  ^
> make[1]: *** [apps/apps.o] Error 1

@native-api  Why is your `Operating system: x86_64-whatever-linux2`? See my output of ./config :
```shell
bash-3.2# cd openssl-OpenSSL_1_1_1c 
bash-3.2# PERL="$HOME/ffmpeg_build/bin/perl" ./config --prefix="$HOME/ffmpeg_build" --openssldir="$HOME/ffmpeg_build" shared zlib 
Operating system: i686-whatever-linux2
Configuring OpenSSL version 1.1.1c (0x1010103fL) for linux-x86
Using os-specific seed configuration
Creating configdata.pm
Creating Makefile

**********************************************************************
***                                                                ***
***   OpenSSL has been successfully configured                     ***
***                                                                ***
***   If you encounter a problem while building, please open an    ***
***   issue on GitHub <https://github.com/openssl/openssl/issues>  ***
***   and include the output from the following command:           ***
***                                                                ***
***       perl configdata.pm --dump                                ***
***                                                                ***
***   (If you are new to OpenSSL, you might want to consult the    ***
***   'Troubleshooting' section in the INSTALL file first)         ***
***                                                                ***
**********************************************************************

```

abhiTronix:
> i686 Dockerfile fails again at `git installation. abhiTronix#3

@native-api and Why are you removing git unnecessarily? It's working all fine here:
```shell
bash-3.2# yum install git -y
Setting up Install Process
Resolving Dependencies
--> Running transaction check
---> Package git.i386 0:1.8.2.3-1.el5 set to be updated
--> Processing Dependency: perl-Git = 1.8.2.3-1.el5 for package: git
--> Processing Dependency: rsync for package: git
--> Processing Dependency: perl(Term::ReadKey) for package: git
--> Processing Dependency: perl(Git) for package: git
--> Processing Dependency: perl(Error) for package: git
--> Processing Dependency: openssh-clients for package: git
--> Running transaction check
---> Package openssh-clients.i386 0:4.3p2-82.el5 set to be updated
--> Processing Dependency: openssh = 4.3p2-82.el5 for package: openssh-clients
--> Processing Dependency: libfipscheck.so.1 for package: openssh-clients
---> Package perl-Error.noarch 1:0.17010-1.el5 set to be updated
---> Package perl-Git.i386 0:1.8.2.3-1.el5 set to be updated
---> Package perl-TermReadKey.i386 0:2.30-4.el5 set to be updated
---> Package rsync.i386 0:3.0.6-6.el5_11 set to be updated
--> Running transaction check
---> Package fipscheck-lib.i386 0:1.2.0-1.el5 set to be updated
--> Processing Dependency: /usr/bin/fipscheck for package: fipscheck-lib
---> Package openssh.i386 0:4.3p2-82.el5 set to be updated
--> Running transaction check
---> Package fipscheck.i386 0:1.2.0-1.el5 set to be updated
--> Finished Dependency Resolution

Dependencies Resolved

============================================================================================================================================================================================================
 Package                                               Arch                                        Version                                               Repository                                    Size
============================================================================================================================================================================================================
Installing:
 git                                                   i386                                        1.8.2.3-1.el5                                         epel                                          16 M
Installing for dependencies:
 fipscheck                                             i386                                        1.2.0-1.el5                                           base                                          13 k
 fipscheck-lib                                         i386                                        1.2.0-1.el5                                           base                                         6.7 k
 openssh                                               i386                                        4.3p2-82.el5                                          base                                         291 k
 openssh-clients                                       i386                                        4.3p2-82.el5                                          base                                         455 k
 perl-Error                                            noarch                                      1:0.17010-1.el5                                       epel                                          26 k
 perl-Git                                              i386                                        1.8.2.3-1.el5                                         epel                                          46 k
 perl-TermReadKey                                      i386                                        2.30-4.el5                                            epel                                          32 k
 rsync                                                 i386                                        3.0.6-6.el5_11                                        updates                                      338 k

Transaction Summary
============================================================================================================================================================================================================
Install       9 Package(s)
Upgrade       0 Package(s)

Total download size: 17 M
Downloading Packages:
(1/9): fipscheck-lib-1.2.0-1.el5.i386.rpm                                                                                                                                            | 6.7 kB     00:00     
(2/9): fipscheck-1.2.0-1.el5.i386.rpm                                                                                                                                                |  13 kB     00:00     
(3/9): perl-Error-0.17010-1.el5.noarch.rpm                                                                                                                                           |  26 kB     00:00     
(4/9): perl-TermReadKey-2.30-4.el5.i386.rpm                                                                                                                                          |  32 kB     00:00     
(5/9): perl-Git-1.8.2.3-1.el5.i386.rpm                                                                                                                                               |  46 kB     00:00     
(6/9): openssh-4.3p2-82.el5.i386.rpm                                                                                                                                                 | 291 kB     00:01     
(7/9): rsync-3.0.6-6.el5_11.i386.rpm                                                                                                                                                 | 338 kB     00:01     
(8/9): openssh-clients-4.3p2-82.el5.i386.rpm                                                                                                                                         | 455 kB     00:01     
(9/9): git-1.8.2.3-1.el5.i386.rpm                                                                                                                                                    |  16 MB     00:45     
------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Total                                                                                                                                                                       317 kB/s |  17 MB     00:55     
Running rpm_check_debug
Running Transaction Test
Finished Transaction Test
Transaction Test Succeeded
Running Transaction
  Installing     : perl-Error                                                                                                                                                                           1/9 
  Installing     : perl-TermReadKey                                                                                                                                                                     2/9 
  Installing     : rsync                                                                                                                                                                                3/9 
  Installing     : fipscheck-lib                                                                                                                                                                        4/9 
  Installing     : openssh                                                                                                                                                                              5/9 
  Installing     : openssh-clients                                                                                                                                                                      6/9 
  Installing     : fipscheck                                                                                                                                                                            7/9 
  Installing     : git                                                                                                                                                                                  8/9 
  Installing     : perl-Git                                                                                                                                                                             9/9 

Installed:
  git.i386 0:1.8.2.3-1.el5                                                                                                                                                                                  

Dependency Installed:
  fipscheck.i386 0:1.2.0-1.el5        fipscheck-lib.i386 0:1.2.0-1.el5  openssh.i386 0:4.3p2-82.el5  openssh-clients.i386 0:4.3p2-82.el5  perl-Error.noarch 1:0.17010-1.el5  perl-Git.i386 0:1.8.2.3-1.el5 
  perl-TermReadKey.i386 0:2.30-4.el5  rsync.i386 0:3.0.6-6.el5_11      

Complete!
```

native-api:
Here's what I'm running (Ubuntu Bionic Desktop x64):
```
$ git clone https://github.com/abhiTronix/opencv-python
$ cd opencv-python/docker/
$ sudo docker build - <Dockerfile_i686
```

I think I know why the commands succeed for you: you are running them in the shell received from `sudo docker run -it quay.io/pypa/manylinux1_i686:latest bash` -- i.e. in the environment before any of the Dockerfile commands have been run, not in the actual environment that this command is run in in `docker build`.
That's why I demand that you `docker build` the dockerfile as a whole before submiting it.

To get a shell in the intermediate environment after a specific command in Dockerfile is run, look at `docker build` output. After each command succeeds, it saves an intermediate image. Supply that image's ID to `docker run` instead of `manylinux1<...>` and you'll get a shell in the actual environment at that point.

native-api:
> ```shell
> bash-3.2# PERL="$HOME/ffmpeg_build/bin/perl" ./config --prefix="$HOME/ffmpeg_build" --openssldir="$HOME/ffmpeg_build" shared zlib 
> Operating system: i686-whatever-linux2
> ```

This actually suggests the reason. If this works for you in a "clean" `manylinux1` environment but not in the environment after running a few commands, then the reason for the breakage must lie in those commands!
In particular, the unnecessary install of `gcc` looks mightly suspicious.
You can e.g. build Perl in multiple intermediate containers and see where it starts breaking.

abhiTronix:
@native-api I think we should wait for @skvark to test these two new dockerfiles with changes. I'll make further improvements if it fails for him too. @skvark kindly post the error message if it fails for you.

abhiTronix:
@native-api Thanks for this changes. Can we finally test this issue [example code](https://github.com/skvark/opencv-python/issues/204#issue-446912322) too? Does this run ok at your end with final build?

skvark:
I'll have a look during the weekend.

native-api:
> Does this run ok at your end with final build?

Rick Roll succeeded

skvark:
I built the images locally and everything went fine. I'll merge this and the other Docker changes before building and uploading the updated images to Quay.

skvark:
Thanks!

PR #231, opened 2019-08-07
Including a custom libjpeg is no longer needed

native-api:
Since https://github.com/opencv/opencv/pull/11497, OpenCV bundles its own.

This code is now unused so no need to keep and maintain it.


native-api:
The suggestion to remove is in https://github.com/skvark/opencv-python/issues/204#issuecomment-516444192

The lib was removed from dockerfiles to resolve a build failure in https://github.com/skvark/opencv-python/pull/229/commits/48085831bcbfcb0e7dbc9e6121cb33d01df76fd2

PR #235, opened 2019-08-17
Allow passing custom CMake arguments

native-api:
Resolves https://github.com/skvark/opencv-python/issues/214


skvark:
Looks good, nice solution.

PR #236, opened 2019-08-17
Don't build unused Qt features; use all cores

native-api:
This drastically reduces Qt compilation time and reduces the resulting image size by about 440M.

The resulting libs included into the wheel are reduced by about 600k.

I've checked that none of the excluded features are used:

* OpenCV only directly links to QtCore, QtGUI and QtTest -- and consequently, these are the only ones included into the manylinux1 wheel by auditwheel.
* Likewise, I checked that all the imports from those used by OpenCV are present in the reduced version.
* There's an issue for me with no icons in Qt `imshow()` UI. But it's also present in the vanilla wheel.
* Also considered `-no-nis` and `-no-cups` but ultimately decided against them: I'm not sure that the UI has no printing capabilities. (NIS support is used solely to discover network printers.)
* Likewise, didn't touch stuff like X features support: there's no such easy way to make sure they are not used.

Merge conflicts with https://github.com/skvark/opencv-python/pull/229 are all but guaranteed.

### DIFF
--- a/docker/Dockerfile_i686
+++ b/docker/Dockerfile_i686
@@ -1,38 +1,38 @@
 FROM quay.io/pypa/manylinux1_i686:latest
 
 RUN curl -O -L https://download.qt.io/archive/qt/4.8/4.8.7/qt-everywhere-opensource-src-4.8.7.tar.gz && \
-	tar -xf qt-everywhere-opensource-src-4.8.7.tar.gz && \
-	cd qt-everywhere* && \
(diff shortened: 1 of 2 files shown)


skvark:
Thank you, that's a quite big change in the image size. I don't think people are using OpenCV to build actual user interfaces so the icon issue is not a big deal. 

There are indeed some merge conflicts which must be resolved before merge.

PR #248, opened 2019-09-19
add multi-jobs build

Borda:
adding an argument for building OpenCV in all available threads

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -663,9 +663,15 @@ before_install: |
     set +x
 
 install: |
+    # for Linux allow parallel  build
+    if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
+        PARALLEL="-j$(nproc)"
+    else
+        PARALLEL=""
+    fi
     # Build and package
     set -x
-    build_wheel $REPO_DIR $PLAT
+    build_wheel $REPO_DIR $PLAT $PARALLEL
     set +x
 
 script: |
--- a/.travis.yml
+++ b/.travis.yml
@@ -663,15 +663,9 @@ before_install: |
     set +x
 
 install: |
-    # for Linux allow parallel  build
-    if [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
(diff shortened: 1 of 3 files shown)


skvark:
Builds are failing. Some issues:

- you have overwritten whole files, please check Git line ending settings
- there's no such thing as ``$(nproc)`` on Windows: `setup.py: error: argument -j: invalid int value: '$(nproc)'`
- macOS and Linux fail with ``setup.py: error: argument -j: expected one argument``

Borda:
I see that it does not work on Windows... but I am surprised that it fails also on Linux since I have built it with this option and it worked fine... I will have look at it...

Borda:
I have set Unix file ending to all project's files but it seems to be wrong. 
@skvark Do you have various file ending depending on running OS?

skvark:
Please check your Git line ending settings: https://stackoverflow.com/a/40821931/5795428

Borda:
@skvark it seems that there is some problem with ffmpeg on xOS:
```
++(travis_config.sh:82): generate_ffmpeg_formula(): git commit -m 'add/update custom ffmpeg 4.2.1'
U	Formula/glib.rb
error: Committing is not possible because you have unmerged files.
hint: Fix them up in the work tree, and then use 'git add/rm <file>'
hint: as appropriate to mark resolution and make a commit.
fatal: Exiting because of an unresolved conflict.
+(travis_config.sh:109): generate_ffmpeg_formula(): shell_session_update
+(multibuild/common_utils.sh:19): shell_session_update(): 
```

skvark:
There was an issue with Travis cache. Additionally, you'll need some changes from master. However, now that I looked at the source,  ``-jN`` is already defined for both Linux and macOS builds in here: https://github.com/skvark/opencv-python/blob/master/travis_config.sh#L22 This applies only in the Travis environment (as do your current changes). 

On local builds you could just export the flag before build in similar manner as above so I'm not sure if this PR is needed at all.

Issue #266, opened 2019-12-07
Deprecated brew tap-pin usage consistently fails CI if cache is not present


mbdevpl:
### Expected behaviour

After forking, I expected the CI to work in the fork.

### Actual behaviour

Builds are consistently failing due to usage of deprecated (see https://brew.sh/2019/04/04/homebrew-2.1.0/ and https://github.com/Homebrew/brew/pull/5925) `brew tap-pin` functionality.

The problem is in line 101 of `travis_config.sh`:

```
brew tap --list-pinned | grep -qxF cartr/qt4 || brew tap-pin cartr/qt4
```

In the main repo, the first part `brew tap --list-pinned | grep -qxF cartr/qt4` always succeeds, I assume because the cache always kicks in.

In the fork however, it always falls through to `brew tap-pin cartr/qt4` because there is no cache, and then it always fails, so cache is never created. If only I could escape this catch 22 ;)

### Steps to reproduce

1. Fork the repo.
2. Enable Travis CI builds for the fork.
3. Trigger a build.
4. See it crash and burn.

Bunch of examples: https://travis-ci.com/mbdevpl/opencv-python/builds

One specific log: https://travis-ci.com/mbdevpl/opencv-python/jobs/263773575 - in this build no changes have been made in the fork whatsoever, it's simply a build of current master branch.

### Solution idea

I guess one quick solution to try would be to use:

```
brew tap-pin cartr/qt4/formula_name_here
```

But unfortunately as I am unfamiliar with the opencv-python CI setup, I have no idea about what should be in the `formula_name_here`. Would any one of maintainers be able to provide ideas to try?

mbdevpl:
And after I solve this in my fork, I'd like to submit a PR so that other new forks don't get stuck on this issue in the future.

mbdevpl:
@skvark Thanks for merging! You might wanna trigger a new build so that cache is fully re-generated because of:

https://travis-ci.org/skvark/opencv-python/jobs/622109414

> Building dependencies took too long. Restart the build in Travis UI to continue from cache.

mbdevpl:
I had click there myself once as well to finish setting up the initial cache. I think somehow the pull request might have triggered a cache rebuild for the main repo...

skvark:
Ah, yes. I cleared the Travis cache because if there's a long time between builds homebrew recipes fail to update sometimes which makes the cache build fail.

PR #267, opened 2019-12-07
Fix for #266 - don't use deprecated brew tap-pin, instead refer to the full package name when installing

mbdevpl:
don't use deprecated brew tap-pin
instead refer to the full package name when installing

fixes https://github.com/skvark/opencv-python/issues/266

context:
https://brew.sh/2019/04/04/homebrew-2.1.0/
https://github.com/Homebrew/brew/pull/5925


skvark:
Thank you!

PR #274, opened 2019-12-27
replace get_config_var()  with get_config_vars() in setup.py.

rockywhisper:
The get_config_var() is deprecated since Python3, and it causes a warning: `warnings.warn('SO is deprecated, use EXT_SUFFIX', DeprecationWarning, 2)`

If we change to `get_config_vars().get("SO")`, it would be compatible with both Python2 and Python3. So why not  :-)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -67,7 +67,7 @@ def main():
 
     package_data = {
         'cv2':
-            ['*%s' % sysconfig.get_config_var('SO')] +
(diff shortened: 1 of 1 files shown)


skvark:
Thanks, I'll wait for the Travis builds to finish before merging (there were some issues with Travis cache so I had to restart the build).

PR #397, opened 2020-09-30
Improve docs

Satyam308:


skvark:
I think that the current spelling is better.

PR #428, opened 2020-12-11
Add docs for debug build

ApproximateIdentity:
This came out of the discussion in this issue: https://github.com/skvark/opencv-python/issues/424

Maybe you'd like to change the presentation here and/or improve the explanation, but this seems to me like a good enough addition.


skvark:
Thanks! Looks fine to me.

Issue #429, opened 2020-12-20
Wheel support for macOS arm64 (Apple Silicon)


richiksc:
OpenCV 4.50 supports building for Apple Silicon (view release notes, PR https://github.com/opencv/opencv/pull/18094 was merged and released in 4.50). Now that the size limit for the project on PyPI has been increased, is there any chance we get prebuilt wheels for macOS arm64? The build script supports cross compilation from x86_64 to arm64, so I don't think dedicated Apple Silicon CI hardware is needed.

skvark:
Everything is pending this largeish CI migration task described in here: https://github.com/skvark/opencv-python/issues/422

This repository does not use those build scripts but it looks like the cross compilation needs just a few additional CMake flags so it shouldn't be a problem to build those wheels in the future.

richiksc:
@skvark Thanks! Looks like Travis CI shutting off OSS repositories has significantly interfered with any work.

skvark:
One important thing that I would like to add to my previous comment: 

It will take some time to get native `numpy` packages for Apple silicon. In practice, this means months, maybe even years. This package depends on `numpy` which means that native `arm64` for Apple silicon won't be possible before a compatible `numpy` version has been released.

Recommended reading: https://github.com/neurolabusc/AppleSiliconForNeuroimaging

richiksc:
@skvark In regards to cross compilation, I don't know whether numpy are using macOS CI runners are not, but Clang 12+ on macOS with the Xcode 12+ Command Line Tools installed supports cross compilation from Intel to `arm64-apple-darwin20.2.0`.

ogrisel:
> It will take some time to get native numpy packages for Apple silicon.

Note that the conda-forge community project had built a M1-native version of the numpy, scipy, pandas, scikit-learn... stack even prior to the release of the first M1 Mac Books. You can install those packages by setting up a conda env with:

https://github.com/conda-forge/miniforge#miniforge3

I use it on a daily basis and it works great. Note that conda-forge does not ship tensorflow though.

skvark:
> > It will take some time to get native numpy packages for Apple silicon.
> 
> Note that the conda-forge community project had built a M1-native version of the numpy, scipy, pandas, scikit-learn... stack even prior to the release of the first M1 Mac Books. You can install those packages by setting up a conda env with:
> 
> https://github.com/conda-forge/miniforge
> 
> I use it on a daily basis and it works great. Note that conda-forge does not ship tensorflow though.

[...]

PR #475, opened 2021-04-30
Make it possible to compile from sources against numpy on macOS aarch64

skvark:
Add rules for minimum `numpy` version for macOS aarch64. This does not solve https://github.com/opencv/opencv-python/issues/429 fully but should make it possible to build both `numpy` and `opencv-python` from sources on M1. See this comment: https://github.com/opencv/opencv-python/issues/429#issuecomment-806069540


skvark:
Later on, the `numpy` version must be updated when they release native wheels.

PR #505, opened 2021-06-26
Fix build for OSX ARM64 - platform says arm64, not aarch64. Also, set…

sergregory:
Fix local build for OS ARM64 as platform reports its name as 'arm64', not 'aarch64'. Also, increased numpy version, to avoid building it from sources.

What needs to be done is a CI step specific for OSX ARM64 - similar to what is done for windows


sergregory:
And do not forget to export flgas: `export LDFLAGS="-L/opt/homebrew/lib"; export CPPFLAGS="-I/opt/homebrew/include`

asenyaev:
Fixes for OSX ARM64 build will be added in another PR.

PR #507, opened 2021-06-30
Added MacOS with M1 build

asenyaev:
1. Created MacOS with M1 build using arm64.
2. Added exceptions for the workflows, which will not trigger other builds if there are no changes in these files.
3. Added exceptions for `arm64` platform in `pyproject.toml`

### DIFF
--- a/.github/workflows/build_wheels_macos_m1.yml
+++ b/.github/workflows/build_wheels_macos_m1.yml
@@ -1,4 +1,4 @@
-name: Build PYPI wheels for opencv-python on Windows
+name: Build PYPI wheels for opencv-python on Macos M1
 
 on:
   pull_request:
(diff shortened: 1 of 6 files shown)


.github/workflows/build_wheels_macos_m1.yml
```diff
@@ -0,0 +1,117 @@
...
+        submodules: false
+        fetch-depth: 0
+
+    - name: Update submodules
+      if: github.event_name == 'pull_request'
+      run: git submodule update --remote
+
+    - name: build script
+      run: |
+        git submodule update --init multibuild
```
sergregory, line 48: We do not need multibuild here, as we do all the build steps manually. To be honest, I've started thinking we do not need multibuild for all other builders too, as our required steps are simple and we already do them for OSX and windows. But this is a discussion for the other PR)

sergregory (approved):

.github/workflows/build_wheels_macos_m1.yml
asenyaev: Okay, I will change the flow of the build without multibuild and prepare another PR.

luisdemarchi:
@sergregory  It's been 14 hours since this merge was done, but it's been 8 hours since I needed to do the manual build. Where did I go wrong?

sergregory:
@luisdemarchi probably, you did everything right - having a merge here doesn't mean we have a new release deployed to PyPI. We try to align the release cycle with the opencv release cycle, so now we're working on 4.5.3 and expect it to be ready this week.

DanLipsitt:
@sergregory Any chance of deploying this first, since right now there's no release for M1 at all?

Issue #571, opened 2021-10-21
Pip build fails: FileNotFoundError: [Errno 2] No such file or directory: 'scripts/__init__.py'


elmeriniemela:
Reproduce:
* `pip install --upgrade pip setuptools wheel`
* `CMAKE_ARGS="-DOPENCV_ENABLE_NONFREE=ON" pip install -v --no-binary=opencv-contrib-python opencv-contrib-python`

Related commit: https://github.com/opencv/opencv-python/commit/6a1f61b8d7e413059dbf0b9e897579c8305c327f#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7R389

Related PR: https://github.com/opencv/opencv-python/pull/563

```
Error:

Configuring Project
    Working directory:
      /tmp/pip-install-ais_cble/opencv-python_36817dc5e0834847b9d8f8f1dc7bd3c4/_skbuild/linux-x86_64-3.8/cmake-build
    Command:
[...]

PR #577, opened 2021-11-01
Fixed a build from the source

asenyaev:
Also fix an issue https://github.com/opencv/opencv-python/issues/571

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -378,8 +378,7 @@ class RearrangeCMakeOutput(object):
         print("Copying files from CMake output")
 
         # add lines from the old __init__.py file to the config file
-        with open('%s/scripts/__init__.py'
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -378,7 +378,8 @@ def _classify_installed_files_override(
         print("Copying files from CMake output")
 
         # add lines from the old __init__.py file to the config file
-        with open('scripts/__init__.py', 'r') as custom_init:
+        with open('%s/scripts/__init__.py'
+        % os.path.split(os.path.abspath(__file__))[0], 'r') as custom_init:
```
asmorkalov, line 382: os.path.join should be better than  "%s"

asenyaev: Added `os.path.join` usage instead of `%s`.

PR #563, opened 2021-10-15
Updated submodules to the release 4.5.4 and added python loader support

asenyaev:

PR #578, opened 2021-11-01
Updated opencv submodule for a tag and the description

asenyaev:
Related to the https://github.com/opencv/opencv-python/pull/563#issuecomment-948846628 about a submodule commit.

Also updated the long description for the python 3.10 support.


asmorkalov (approved):
:+1:

Issue #570, opened 2021-10-21
Pylint no-member warnings in 4.5.4.58


robertschweizer:
### Expected behaviour

Pylint was previously able to inspect `cv2` member existence with the `extension-pkg-whitelist` parameter.

### Actual behaviour

For opencv-python-4.5.4.58, pylint can no longer inspect the members.

### Steps to reproduce

On Windows 10, Python 3.8 64bit, using Pylint 2.11.1:

```py
"""Test module test.py"""
from cv2 import getStructuringElement, MORPH_ELLIPSE

a = getStructuringElement(MORPH_ELLIPSE, (3, 3))
```

With opencv-python-4.5.3.56:

```
$ pylint test.py --extension-pkg-whitelist=cv2

-------------------------------------------------------------------
Your code has been rated at 10.00/10 (previous run: 5.00/10, +5.00)
```

With opencv-python-4.5.4.58:

```
$ pylint test.py --extension-pkg-whitelist=cv2
************* Module test
test.py:2:0: E0611: No name 'getStructuringElement' in module 'cv2' (no-name-in-module)
test.py:2:0: E0611: No name 'MORPH_ELLIPSE' in module 'cv2' (no-name-in-module)

----------------------------------------------------------------------
Your code has been rated at -40.00/10 (previous run: 10.00/10, -50.00)
```

##### Issue submission checklist

 - [x] This is not a generic OpenCV usage question (looking for help for coding, other usage questions, homework etc.)
   
 - [x] I have read the README of this repository and understand that this repository provides only an automated build toolchain for OpenCV Python packages (there is no actual OpenCV code here)
   
 - [x] The issue is related to the build scripts in this repository, to the pre-built binaries or is a feature request (such as "please enable this additional dependency")
 - [x] I'm using the latest version of ``opencv-python``

robertschweizer:
Good idea, maybe the C extension name just changed. But no, with `--unsafe-load-any-extension=y` it also does not work.

asmorkalov:
Since 4.5.4.60 pylint works with OpenCV python loader correctly. Command line example:
```
python3 -m pylint --extension-pkg-whitelist=cv2,numpy ./contours.py
```

robertschweizer:
The fix for this was reverted in 5633cf319efadb628e8369b48834d74491c10b1a, so this bug appears again in version 4.6.0.66. Should we rename and reopen this issue or create a new one?

pasbi:
[...]

PR #579, opened 2021-11-03
Added lines for a proper work using pylint and an autocomplete in IDE

asenyaev:
PR also fixes https://github.com/opencv/opencv-python/issues/570 and https://youtrack.jetbrains.com/issue/PY-35691

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -377,6 +377,22 @@ class RearrangeCMakeOutput(object):
 
         print("Copying files from CMake output")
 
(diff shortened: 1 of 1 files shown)


asmorkalov (approved):
os.path.join is better than substitution

asenyaev:
> os.path.join is better than substitution

Changed it in the latest commit.

asmorkalov (approved):

PR #580, opened 2021-11-03
Use PEP-508 compliant dependency specifiers

jreese:
This allows correctly resolving the numpy version requirement via PyPI metadata, using the same metadata regardless of what version of Python the package was built with or which wheel is uploaded to PyPI first.

PEP 508: https://www.python.org/dev/peps/pep-0508/


asenyaev (approved):

asmorkalov (approved):
:+1:

PR #584, opened 2021-11-11
Update 3.4 branch to the latest changes in master

asenyaev:
Tested in my fork repo:
- [macOS](https://github.com/asenyaev/opencv-python/actions/runs/1448736514)
- [Ubuntu](https://github.com/asenyaev/opencv-python/actions/runs/1448736515)
- [Windows](https://github.com/asenyaev/opencv-python/actions/runs/1448736513)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -366,15 +366,13 @@ class RearrangeCMakeOutput(object):
         print("Copying files from CMake output")
 
         # lines for a proper work using pylint and an autocomplete in IDE
-        with open("%spython/cv2/__init__.py"
-        % cmake_install_dir, 'r') as opencv_init:
+        with open(os.path.join(cmake_install_dir, "python", "cv2", "__init__.py"), 'r') as opencv_init:
             opencv_init_lines = opencv_init.readlines()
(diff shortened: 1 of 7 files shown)


setup.py
```diff
@@ -365,8 +365,20 @@ def _classify_installed_files_override(
...
+        with open("%spython/cv2/__init__.py"
+        % cmake_install_dir, 'r') as opencv_init:
+            opencv_init_lines = opencv_init.readlines()
+            extra_imports = ('\nfrom .cv2 import *\nfrom . import data\n')
+            free_line_after_imports = 5
+            opencv_init_lines.insert(free_line_after_imports, extra_imports)
+
+        with open("%spython/cv2/__init__.py"
+        % cmake_install_dir, 'w') as opencv_final_init:
+            opencv_final_init.writelines(opencv_init_lines)
```
asenyaev, line 378: The same logic as in https://github.com/opencv/opencv-python/pull/579, but we have here only two imports.

asmorkalov: os.path.join should be better than string substitution.

asenyaev: Updated it in the latest commit

asmorkalov (approved):

PR #589, opened 2021-11-26
Fix some but not all undefined names in scripts/__init__.py

cclauss:
Add missing imports that were removed in #563

### DIFF
--- a/scripts/__init__.py
+++ b/scripts/__init__.py
@@ -1,3 +1,6 @@
+import os
+import sys
+
 PYTHON_EXTENSIONS_PATHS = [
     LOADER_DIR
 ] + PYTHON_EXTENSIONS_PATHS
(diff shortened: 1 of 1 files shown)


scripts/__init__.py
```diff
@@ -1,3 +1,6 @@
+import os
+import sys
```
alalek, line 2: > scripts/__init__.py

AFAIK, it is not a standalone script. It is used through eval()

cclauss: [eval() is evil](https://nedbatchelder.com/blog/201206/eval_really_is_dangerous.html) ;-) but if these imports have already been done then these lines are no-ops but if the imports have not been done then these lines will prevent crashes.

asenyaev: @cclauss, you mean it prevents crashes in a linting job?

cclauss: Crashes in any situation where `os` and `sys` have not already been imported by some other file.

As it says in The Zen of Python", _explicit is better than implicit._

PR #592, opened 2021-12-07
Moved binary from the package root directory to python-X.Y

asenyaev:
- Moved a binary (`.so` / `.pyd`) from `cv2/` folder in a package to `cv2/python-X.Y`.
- Added an overwriting for `config.py` file instead of `config-X.Y.py`, because we have a proper path to our binary in `config-X.Y.py` file, but `config.py` doesn't have a proper path to other binaries (e.g. the path is to `/lib64` in a docker container)
- Added 3.4 branch for builds in Pull Requests in case of we have different `setup.py` files

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -23,7 +23,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest]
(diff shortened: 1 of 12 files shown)


asenyaev:
I will update `opencv` submodule after merging https://github.com/opencv/opencv/pull/21209 to start actions.

PR #595, opened 2021-12-15
Added python limited API support

asenyaev:
- Added `"DPYTHON3_LIMITED_API=ON"` flag in `setup.py`
- Changed the pattern to save a binary in `setup.py`
- Added a proper import `version.py` in `setup.py` after moving to python limited API
- Changed the command to build a package in `travis_config.sh`
- Modified workflows for all platforms to build packages on one python version and then to test it on all combinations.

Also PR fixes an issue: https://github.com/opencv/opencv-python/issues/414

PR #596, opened 2021-12-15
Changed a path to config-X.py file after python limited API support, optimized workflows

asenyaev:
- Forgot to update a path to `config-X.py` file in `setup.py` after implementing python limited API support in https://github.com/opencv/opencv-python/pull/595, which luckily didn't broke a package, but it breaks for 3.4 branch.
- Optimized uploading and downloading artifacts in workflows.


asmorkalov (approved):
:+1:

PR #635, opened 2022-03-15
Update LICENSE.txt

DavidChinedu23:
I just wanted to update the license, there was no date given, so I thought 2022 might be a good idea.


skvark:
https://hynek.me/til/copyright-years/

PR #638, opened 2022-03-22
Added pylint test

asenyaev:


.github/workflows/build_wheels_macos.yml
```diff
@@ -171,6 +171,11 @@ jobs:
         cd ${{ github.workspace }}/opencv
         python modules/python/test/test.py -v --repo .
 
+    - name: Pylint test
+      run: |
+        python -m pip install pylint
```
asmorkalov, line 176: I propose to pin pylint version

tests/pylint.py
```diff
@@ -0,0 +1,3 @@
+from cv2 import getStructuringElement, MORPH_ELLIPSE
+
+a = getStructuringElement(MORPH_ELLIPSE, (3, 3))
```
asmorkalov, line 3: Why do you need it? I propose to use one of OpenCV samples, if you need entry point.

travis_config.sh
```diff
@@ -141,37 +141,36 @@ function run_tests {
...
 
     #Test package
     $PYTHON modules/python/test/test.py -v --repo .
 }
 
+function pylint_test {
+
+    echo "Starting Pylint tests..."
+
+    $PYTHON -m pip install pylint
```
asmorkalov, line 171: Pin version.

.github/workflows/build_wheels_macos.yml
asenyaev: Pinned `2.12.2` version.

travis_config.sh
asenyaev: Pinned `2.12.2` version.

tests/pylint.py
asenyaev: Tests in `opencv/modules/python/test/` doesn't pass a pylint test, there are some issues.

asmorkalov: https://pullrequest.opencv.org/buildbot/builders/precommit_docs/builds/34604/steps/Run%20check_pylint/logs/stdio

asenyaev: Currently, we cannot do the same logic using `cmake`, because the test part doesn't have the `build` folder in opencv.

Added `pylintrc` in this repository from `opencv/platforms/scripts/pylintrc` (didn't provide it via flag, because the current content won't show any output for our test).

Btw, changed the test file from a custom script to an opencv sample (`opencv/samples/python/squares.py`).

asmorkalov (approved):

Issue #636, opened 2022-03-16
Build fails "late" if no internet connection


GuillaumeDesforges:
##### System information (version)

- OpenCV => 4.5.5.64
- Operating System / Platform => x86_64-linux
- Compiler => 

##### Detailed description

If I clone the repository and run the build in an isolated environment without internet, the build first fails to download the GAPI component

https://github.com/opencv/opencv/blob/e0ffd3e8a5c25b634a853ce695c4d23d3057b361/modules/gapi/cmake/DownloadADE.cmake#L5-L14

and that failure does not stop the build.

However, at the end of the build, it fails to build the wheel

```
  Traceback (most recent call last):
    File "/nix/store/yfwbxnn1j32nyfhi2yzx859dij4xgl3s-python3.9-pip-21.3.1/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 363, in <module>
      main()
    File "/nix/store/yfwbxnn1j32nyfhi2yzx859dij4xgl3s-python3.9-pip-21.3.1/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 345, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "/nix/store/yfwbxnn1j32nyfhi2yzx859dij4xgl3s-python3.9-pip-21.3.1/lib/python3.9/site-packages/pip/_vendor/pep517/in_process/_in_process.py", line 261, in build_wheel
      return _build_backend().build_wheel(wheel_directory, config_settings,
    File "/nix/store/0xxs98jm9dpvybfgh7kghdq8ayvkp87m-python3.9-setuptools-57.2.0/lib/python3.9/site-packages/setuptools/build_meta.py", line 221, in build_wheel
      return self._build_with_temp_dir(['bdist_wheel'], '.whl',
    File "/nix/store/0xxs98jm9dpvybfgh7kghdq8ayvkp87m-python3.9-setuptools-57.2.0/lib/python3.9/site-packages/setuptools/build_meta.py", line 207, in _build_with_temp_dir
      self.run_setup()
    File "/nix/store/0xxs98jm9dpvybfgh7kghdq8ayvkp87m-python3.9-setuptools-57.2.0/lib/python3.9/site-packages/setuptools/build_meta.py", line 258, in run_setup
      super(_BuildMetaLegacyBackend,
[...]

PR #639, opened 2022-03-24
Added pre-downloaded 3rdparty files to source packages

asenyaev:
This PR fixing https://github.com/opencv/opencv-python/issues/636


asmorkalov (approved):

alalek:
BTW, set of external files are different. It may depend on OS / 64/32-bit builds.
AFAIK, we have one "source" package.

PR #652, opened 2022-04-07
Pinned versions for dependencies

asenyaev:


pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
 requires = [
-  "setuptools", "wheel", "cmake", "pip",
+  "setuptools==59.2.0", "wheel==0.37.0", "cmake==3.22.3", "pip",
```
asmorkalov, line 3: CMake version is too hight. Regular Ubuntu user cannot rebuild the package with system cmake. I propose to have `>= 3.1` for CMake as they presume compatibility carefully.

asmorkalov (approved):

pyproject.toml
asenyaev: Updated cmake version to `3.6.3`, because it's the lowest available version in PyPi.

asmorkalov: Ubuntu 16.04 provides cmake 3.5.1-1ubuntu3. I think it should work for us.

asenyaev: Okay, defined `3.1`.

asmorkalov (approved):

PR #681, opened 2022-06-23
Changed labeles for macOS ARM64 and Linux ARM64 builds

asenyaev:


fengyuentau:
It seems a lot of workflows are added to the machines on my side. Will it be too heavy for them?

asenyaev:
> It seems a lot of workflows are added to the machines on my side. Will it be too heavy for them?

No, there are only 4 builds for each machine and other workflows are OpenCV python tests. In addition, this repository runs workflows quite rare.

asenyaev:
The last one issue - the network connection of Linux ARM64 machine.

fengyuentau (approved):
Some of the cleanup steps use `asenyaev/cleanup-action@v1`, some others use `find . -mindepth 1 -delete`, and the others use ` rm -rf ./*`. If your `asenyaev/cleanup-action@v1` can work across platforms, then just use your cleanup action for simplicity and unity.

BTW, Zihao is working on the network issue on the Linux ARM64 machine. It seems the problem is somehow related to the IPV6 routing on the machine.

Others look good to me👍

asenyaev:
@fengyuentau, thank you for the update!

My custom action works only with docker, because it was created using dockerfile and entrypoint bash script. I'm going to rebuild docker images which are used for build jobs using the non-root user to avoid permission issues when need to delete also hidden files (`.git`).

asmorkalov (approved):

PR #684, opened 2022-06-24
Updated multibuild to upgrade default 64-bit test image to focal

radarhere:
Updates multibuild to include https://github.com/multi-build/multibuild/pull/433

Also updates the multibuild repository URL, as it now lives at https://github.com/multi-build/multibuild/


asenyaev (approved):
Thank you! :+1:

PR #699, opened 2022-08-02
Scheduled rolling workflows

asenyaev:


asenyaev:
Builds with forced `ENABLE_ROLLING: 1` are here:
- [Linux x86_64](https://github.com/opencv/opencv-python/actions/runs/2782154395)
- [Linux ARM64](https://github.com/opencv/opencv-python/actions/runs/2782154390)
- [Windows x86 / x64](https://github.com/opencv/opencv-python/actions/runs/2782154393) - there are issues after merging https://github.com/opencv/opencv/pull/22196
- [macOS x86_64](https://github.com/opencv/opencv-python/actions/runs/2782154394)
- [macOS ARM64](https://github.com/opencv/opencv-python/actions/runs/2782154391)

.github/workflows/build_wheels_windows.yml
```diff
@@ -1,4 +1,4 @@
-name: Windows x86/x64
+name: Windows x86_x64
```
fengyuentau: I think it should be `x86_64` if putting together.

fengyuentau (approved):
👍

asmorkalov (approved):
:+1:

.github/workflows/build_wheels_windows.yml
asenyaev: Edited typing.

asenyaev:
macOS x86_64 are cancelled because they run on macos-10.15 GitHub Runner which is deprecated now and going to be removed (https://github.com/actions/virtual-environments/issues/5583).

I will merge this PR when other pipelines are passed and going to change the runner for these builds. Btw, macOS x86_64 pipelines are passed in a previous commit ([pipelines](https://github.com/opencv/opencv-python/runs/7640798543?check_suite_focus=true)), the last commit didn't affect it.

fengyuentau:
Kindly ask when these packages will be ready for users to download using `pip install opencv-python-rolling`? I tested just now and it was not available.

asenyaev:
Yes, these packages are not released yet. They are scheduled for a weekly release on Saturday.

fengyuentau:
Is there a data when they will be released?

asenyaev:
@fengyuentau they are released now. The schedule is configured to run it on Saturday at 3 am (UTC +0)
```
  schedule:
    - cron: '0 3 * * 6'
```
I see that only source packages with Contrib hasn't been released in case of PyPi limits, the default size is 100Gb, but these packages are about 150Gb. Need to request for an extension.

fengyuentau:
Thanks for clarification. I did read the cron schedule and those pkgs were not available earlier today, but now they are working great!

PR #715, opened 2022-08-25
Build zlib from the source on Linux with the latest version

asenyaev:

PR #718, opened 2022-08-26
Build zlib from the source on Linux with the latest version (#715)

asenyaev:
* Added zlib build from the source on Linux, handling auditwheel issues

* Removed unused variable

* Removed bzip2 from dependencies for OpenCV Python packages


asmorkalov (approved):
:+1:

PR #751, opened 2022-11-29
OpenCV dependnecies update.

asmorkalov:


alalek (approved):
LGTM :+1:

But need to repair CI.

asmorkalov:
Looks like CMake issue: https://stackoverflow.com/questions/74162633/problem-compiling-from-source-opencv-with-mvsc2019-in-64-bit-version

PR #752, opened 2022-12-01
As/windws d size

asmorkalov:


asmorkalov:
Looks like CMake issue: https://stackoverflow.com/questions/74162633/problem-compiling-from-source-opencv-with-mvsc2019-in-64-bit-version

PR #758, opened 2022-12-12
Use local pypi mirror for self-hosted runners

asmorkalov:


fengyuentau:
Are these packages built in docker containers? If no, directly setting mirror for pip per host OS can be much easier.

asenyaev:
@fengyuentau they build inside a docker container. Moreover, Linux ARM64 build runs a container using submodule commands what is inconvenient.

One more thing, if you set a configuration for PyPi on your host (macOS for example), we need to configure pypi.org for release stage anyway. But it can prevent any issues with Python packages in other pipelines/jobs/steps.

fengyuentau:
@asenyaev Thank you for the clarification.

Do we have networking issues uploading built packages for release onto pypi.org as well? I did not consider this. The mirror solution is proposed to solve any downloading issues.

asenyaev:
@fengyuentau No, there is GitHub hosted runner for the release stage. Yeah, in this case it’s better to define default PyPi url globally on OpenCV CN runners.

asmorkalov:
Replaced by https://github.com/opencv/opencv-python/pull/759

PR #759, opened 2022-12-12
Update pypi default index url

asenyaev:


asenyaev:
cc @asmorkalov @fengyuentau 

I created this branch to test your idea. I just only see that macOS ARM64 runner is not up, but this should work there also.

asenyaev:
Btw, there is something wrong with Linux ARM64 runners (1, 3 agents). They cannot clone submodules using git from `setup.py`.

Logs:
1. https://github.com/opencv/opencv-python/actions/runs/3678448770/jobs/6221822442#step:5:171
3. https://github.com/opencv/opencv-python/actions/runs/3678448770/jobs/6221822599#step:5:171

Looks like need to clean local cached folders.

asmorkalov (approved):
Thanks a lot for the contribution!

fengyuentau (approved):
👍 Thank you!

Issue #772, opened 2023-01-04
Can not found zlib 1.2.9


wqh17101:
### System Information

OpenCV python version: 4.7.0.68 & 4.6.0.66
Operating System / Platform: Centos 7
Python version: 3.9.2

### Detailed description

I install opencv-python==4.7.0.68 and got 
```
[2023/01/04 10:30:57.244] Traceback (most recent call last):
[2023/01/04 10:30:57.244]   File "/home/xxx/source_code/xxx/OCR/pse/test.py", line 7, in <module>
[2023/01/04 10:30:57.244]     import cv2
[2023/01/04 10:30:57.244]   File "/usr/local/python3.9/lib/python3.9/site-packages/cv2/__init__.py", line 181, in <module>
[2023/01/04 10:30:57.244]     bootstrap()
[2023/01/04 10:30:57.244]   File "/usr/local/python3.9/lib/python3.9/site-packages/cv2/__init__.py", line 153, in bootstrap
[2023/01/04 10:30:57.244]     native_module = importlib.import_module("cv2")
[2023/01/04 10:30:57.244]   File "/usr/local/python3.9/lib/python3.9/importlib/__init__.py", line 127, in import_module
[2023/01/04 10:30:57.244]     return _bootstrap._gcd_import(name[level:], package, level)
[2023/01/04 10:30:57.244] ImportError: /lib64/libz.so.1: version `ZLIB_1.2.9' not found (required by /usr/local/python3.9/lib/python3.9/site-packages/cv2/../opencv_python.libs/libpng16-186fce2e.so.16.37.0)
```
However the same environment，it works well with opencv-python==4.6.0.66

### Steps to reproduce

pip install opencv-python==4.7.0.68
and run `import cv2`

### Issue submission checklist

- [X] I report the issue, it's not a question
- [X] I checked the problem with documentation, FAQ, open issues, forum.opencv.org, Stack Overflow, etc and have not found any solution
- [X] I updated to the latest OpenCV version and the issue is still there
- [X] There is reproducer code and related data files (videos, images, onnx, etc)

Maxlinn:
come here by seaching for solutions, same issue.

i definitely believe this issue was introduced in no more than 7 days, because a program i used very often will always fetch the latest version of `opencv-python` via pip, it was fine last week.

Maxlinn:
> come here by seaching for solutions, same issue.
> 
[...]

PR #780, opened 2023-01-17
OpenCV package does not distribute zlib

asmorkalov:
Roll back to manylinux version to presume binary compatibility.
Related changes: https://github.com/opencv/opencv-python/pull/715
Address:
- https://github.com/opencv/opencv-python/issues/765
- https://github.com/opencv/opencv-python/issues/772


asenyaev (approved):
👍

PR #816, opened 2023-03-10
Build and release wheels for all supported python versions

JacobHayes:
The PyPI releases only have wheels for python 3.7, but 3.7-3.11 are supported and tested. This updates CI to build wheels for all of these versions, which should ease installation for users (and may help some of the other installation issues).

### DIFF
--- a/.github/workflows/build_wheels_linux.yml
+++ b/.github/workflows/build_wheels_linux.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.7']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
         platform: [x64]
         with_contrib: [0, 1]
         without_gui: [0, 1]
--- a/.github/workflows/build_wheels_linux_arm.yml
+++ b/.github/workflows/build_wheels_linux_arm.yml
@@ -27,7 +27,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.7']
+        python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
         platform: [x64]
         with_contrib: [0, 1]
         without_gui: [0, 1]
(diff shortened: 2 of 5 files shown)


skvark:
You don't need to create a wheel for all versions. The released wheels are compatible with newer versions of Python as well due to abi3-compatibility: https://docs.python.org/3/c-api/stable.html

They will install and work in Python 3.7 and all versions above that.

JacobHayes:
Hmm, I think you're right, but it looks like the newer x86-64 wheels were built with a newer macos version, so if you're on an older macos version, it'll pull the `.tar.gz` (or an old opencv version, if you force installing from wheels), eg:
```
$ pip3 install opencv-python-headless
Collecting opencv-python-headless
  Downloading opencv-python-headless-4.7.0.72.tar.gz (91.1 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0.8/91.1 MB 498.0 kB/s eta 0:03:02
ERROR: Operation cancelled by user
$ pip3 install opencv-python-headless --only-binary=:all:
Collecting opencv-python-headless
  Downloading opencv_python_headless-4.7.0.68-cp37-abi3-macosx_10_13_x86_64.whl (51.7 MB)
     ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 51.7/51.7 MB 14.5 MB/s eta 0:00:00
Requirement already satisfied: numpy>=1.17.3 in /Users/jamesoswald/.virtualenvs/dyno-cbd5d54a-py3.10.9/lib/python3.10/site-packages (from opencv-python-headless) (1.23.5)
Installing collected packages: opencv-python-headless
Successfully installed opencv-python-headless-4.7.0.68
```
Eg: v...68 for x86_64 uses 10_13
Screenshot 2023-03-10 at 14 02 25
while v...72 for x86_64 uses 10_13
Screenshot 2023-03-10 at 14 02 31

---
Happy to close this since it should probably be fixed as folks upgrade mac versions.

skvark:
Yes, that's documented in the README file. GitHub Actions deprecated support for MacOS 10, so only version 11 and above is supported in the latest releases.

PR #831, opened 2023-04-11
fix: update build system with CMAKE_ARGS from scikit-build

### LINKED ISSUE #830: config: CMAKE_ARGS already supported by scikit-build
The setup.py in OpenCV-Python manually checks for `CMAKE_ARGS`, and places that in the `cmake_args=` setup keyword. Scikit-build itself does this, and also does some post-processing on CMAKE_ARGS that is missing here, including stripping `CMAKE_INSTALL_PREFIX`, which cannot be set by projects. The recent (a few hours ago) 0.17.0 release of scikit-build now produces an error if it detects CMAKE_INSTALL_PREFIX in `cmake_args=` (since it should not be passed in), but it ignores CMAKE_ARGS (since that get set lots of places like Conda).

The ideal fix would be to require scikit-build 0.14+, another fix would be to duplicate the filtering code in scikit-build: https://github.com/scikit-build/scikit-build/blob/6df6d571ba5c7be03729dd6b4d5e543c8aaae781/skbuild/setuptools_wrap.py#L545.

Seen in https://github.com/pyodide/pyodide/issues/3753. That was easy to fix, but it's likely going to hit (at least) conda builds.

##### Issue submission checklist

 - [x] This is not a generic OpenCV usage question (looking for help for coding, other usage questions, homework etc.)
   
 - [x] I have read the README of this repository and understand that this repository provides only an automated build toolchain for OpenCV Python packages (there is no actual OpenCV code here)
   
 - [x] The issue is related to the build scripts in this repository, to the pre-built binaries or is a feature request (such as "please enable this additional dependency")
 - [x] I'm using the latest version of ``opencv-python``

henryiii:
Fix #830. Touched up a few things related to the build setup.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,7 +1,6 @@
 [build-system]
 requires = [
-  "setuptools==59.2.0", "wheel==0.37.0", "cmake>=3.1", "pip",
-  "scikit-build>=0.13.2",
+  "cmake>=3.1",
   "numpy==1.13.3; python_version=='3.6' and platform_machine != 'aarch64' and platform_machine != 'arm64'",
   "numpy==1.17.0; python_version=='3.7' and platform_machine != 'aarch64' and platform_machine != 'arm64'",
   "numpy==1.17.5; python_version=='3.8' and platform_machine != 'aarch64' and platform_machine != 'arm64'",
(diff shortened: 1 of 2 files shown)


henryiii:
I've reverted the oldest-supported-numpy change to see if that fixes it though I think it was very suspicious that one single file failed before with what looked like missing Python headers - that doesn't look like a problem I caused. :)

Hmm, #837 mentions 1.17.3 fails now. Is it because it needs the Python header included in this case? Anyway, it's easy to override one case while still using oldest-supported-numpy.

asmorkalov:
@henryiii Thanks a lot for the patch! I squashed commits and rebased to  4.x to resolve conflicts.

asmorkalov (approved):
:+1:

PR #862, opened 2023-06-28
Build environment update on Linux.

asmorkalov:


opencv-alalek (approved):
LGTM :+1:

(looks like, CI is red due to network issues)

Issue #825, opened 2023-03-30
pip install build issue


DrBwts:
### Expected behaviour
I expected OpenCV's python API to install. I have used the same proceedure in numerous conda environments & have never had a problem like this. In fact it worked just fine yesterday in a different conda env

### Actual behaviour
`pip` installer threw pages & pages of errors, c&p below

### Steps to reproduce
As admin,
`conda activate genicam`

Then any of the following,

`pip install opencv-contrib-python`
`pip install opencv-contrib-python --user`
`pip install opencv-python`
`pip install opencv-python --user`

They all give the same pages & pages of errors

- operating system - Win 10
- architecture (e.g. x86) - x64
- opencv-python version - 4.7.0.72

I would post the errors but "There was an error creating your Issue: body is too long (maximum is 65536 characters). "

So some highlights, 
```
ERROR: Command errored out with exit status 1:
   command: 'C:\ProgramData\Anaconda3\envs\genicam\python.exe' 'C:\ProgramData\Anaconda3\envs\genicam\lib\site-packages\pip\_vendor\pep517\in_process\_in_process.py' build_wheel 'C:\Users\NICBWT~1\AppData\Local\Temp\tmp98_vz8u5'
       cwd: C:\Users\NICBWT~1\AppData\Local\Temp\pip-install-71z9rl7t\opencv-contrib-python_387a51900d48422f8ed4ed222402118d
  Complete output (1205 lines):


  --------------------------------------------------------------------------------
  -- Trying 'Ninja (Visual Studio 17 2022 x64 v143)' generator
  --------------------------------
  ---------------------------
  ----------------------
  -----------------
  ------------
  -------
  --
  Not searching for unused variables given on the command line.
  CMake Error at CMakeLists.txt:2 (PROJECT):
    Generator

      Ninja

    does not support platform specification, but platform

      x64

    was specified.

The LOADS of stuff ending with...

ERROR: Could not build wheels for opencv-contrib-python, which is required to install pyproject.toml-based projects
```

asmorkalov:
The logs piece says that pip tries to build the package from source, but does not use pre-built wheel. What is your python version? Python 3.6 was deprecated some time ago. 4.7.0.x was built for Python 3.7+.

henryiii:
FYI, it's better to drop 3.6 than to keep supporting it but not provide binaries. If you drop it, then Python 3.6 remain happy with their previously built binaries.

If you are on conda, you should be using conda to get most packages, I think.

[...]

PR #863, opened 2023-06-29
Do not force CMAKE_GENERATOR_PLATFORM on Windows x86_64

asmorkalov:
Tries to address https://github.com/opencv/opencv-python/issues/825


asmorkalov:
cc @henryiii

PR #873, opened 2023-07-13
Fix using only one CPU core when compiling with Unix Makefiles

tangtang666:


setup.py
```diff
@@ -9,11 +9,12 @@
 import platform
 import skbuild
 from skbuild import cmaker
+import multiprocessing
 
 
 def main():
     os.chdir(os.path.dirname(os.path.abspath(__file__)))
-
+    parallel_level = int(multiprocessing.cpu_count() / 2)
```
opencv-alalek: multiprocessing.cpu_count() = 1

then

parallel_level = 0

setup.py
```diff
@@ -222,6 +223,9 @@ def main():
 
     if sys.platform.startswith("linux") and not is64 and "bdist_wheel" in sys.argv:
         subprocess.check_call("patch -p0 < patches/patchOpenEXR", shell=True)
+    if sys.platform.startswith("linux"):
+        # add environ MAKEFLAGS = cpu_core_num/2
+        os.environ['MAKEFLAGS'] = '-j ' + str(parallel_level)
```
opencv-alalek, line 229: What if there is already `MAKEFLAGS` with own value?

PR #904, opened 2023-10-03
Python 3.12 support

asmorkalov:


asmorkalov:
@opencv-alalek Could you take a look? I'm working on Mac builds and will submit separate PR.

opencv-alalek (approved):

PR #936, opened 2023-12-26
Mac Intel refresh

asmorkalov:


opencv-alalek (approved):
Overall looks good to me.

BTW, OpenCV is 4.8.1.

PR #968, opened 2024-03-26
language edit

henrykironde:


rotu:
The primary branch is named `master`. I don't think this change should be made without a corresponding change to the repo and [other places it's referenced in this repo](https://github.com/search?q=repo%3Aopencv%2Fopencv-python%20%20master&type=code).

PR #1001, opened 2024-06-16
WIP: Use Numpy 2.0 for Python 3.9+ packages

asmorkalov:
Numpy 2.0 is backward compatible with 1.x, but introduce several API and ABI changes. Numpy authors claim, that packages built against Numpy 2.0 should work with 1.x too. The PR introduce additional packages built with Python 3.9 and Numpy 2.0. Original packages for Python 3.7+ and Numpy 1.x presumed for compatibility. 

Numpy release details: https://github.com/numpy/numpy/releases/tag/v2.0.0

PR #1002, opened 2024-06-17
Use Numpy 2.0 for Python 3.9+ packages

asmorkalov:
Numpy 2.0 is backward compatible with 1.x, but introduce several API and ABI changes. Numpy authors claim, that packages built against Numpy 2.0 should work with 1.x too. The PR introduce additional packages built with Python 3.9 and Numpy 2.0.

Numpy release details: https://github.com/numpy/numpy/releases/tag/v2.0.0

The PR differs from https://github.com/opencv/opencv-python/pull/1001. It builds packages with limited API cp37, but with Python 3.9 and Numpy 2.0. It's expected that the package should work with all supported Pythons 3.7+ and Numpy 1.x and 2.x.


pyproject.toml
```diff
@@ -4,13 +4,9 @@ requires = [
   "numpy==1.13.3; python_version=='3.6' and platform_machine != 'aarch64' and platform_machine != 'arm64'",
   "numpy==1.17.0; python_version=='3.7' and platform_machine != 'aarch64' and platform_machine != 'arm64'",
   "numpy==1.17.5; python_version=='3.8' and platform_machine != 'aarch64' and platform_machine != 'arm64'",
-  "numpy==1.19.3; python_version<='3.9' and sys_platform == 'linux' and platform_machine == 'aarch64'",
```
opencv-alalek, line 7: We need explicit `<2` in `setup.py` for Python 3.8 and earlier in case of possible expanding of numpy2 releases.

rgommers: It would clarify the actual state, but the current version works as well since numpy itself has `python_requires>=3.9`, so numpy 2.0 and later cannot be installed with older Python versions.

The added constraints here look good to me.

sergregory (approved):
LGTM

caldempsey:
🙏

PR #1003, opened 2024-06-17
Until NumPy 2.0 is supported, it might be nice to specify numpy<2 in setup.py

sibocw:
Until a longer term solution is found to make opencv-python compatible with numpy 2.0, it might be nice to add `numpy>=xxx,<2` in setup.py. This can facilitate, for example, installation via pip and CI/CD in downstream libraries that depend opencv-python.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,16 +23,16 @@ def main():
     build_rolling = get_build_env_var_by_name("rolling")
 
     install_requires = [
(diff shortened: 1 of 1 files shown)


opencv-alalek:
#1002

asmorkalov:
Closed in favor of https://github.com/opencv/opencv-python/pull/1002

PR #1061, opened 2024-12-12
3.12 functionality

SaikWolf:
Since the depreciation has been coming for a while, and 3.12 removes it, why not condition the changes from 3.12 onwards?


asmorkalov (approved):

twistedpair:
Perfect timing as I've been running into this. 
When will the version bump push to PyPi?

zhanwenchen:
> Perfect timing as I've been running into this. When will the version bump push to PyPi?

It's live on PyPI 4 hours ago: https://pypi.org/project/opencv-python/#history

PR #1100, opened 2025-05-13
CI update

avdivan:


setup.py
```diff
@@ -193,6 +193,7 @@ def main():
             "-DBUILD_DOCS=OFF",
             "-DPYTHON3_LIMITED_API=ON",
             "-DBUILD_OPENEXR=ON",
+            "-DWITH_MSMF=OFF",
```
asmorkalov: It should break regular opencv-python build. Let's exclude it from the PR now and merge other CI related changes.

asmorkalov (approved):

PR #1103, opened 2025-06-05
Visualization of the opencv-python codebase

ivanmilevtues:
In this PR I have generated high-level diagram to help new developers to get to know the opencv-python project (from the stand point of contributors).

Me and a friend of mine are on a quest to help people get to know codebases faster. Thus we think visualization is very important and decided to generate high-level diagrams as we think this is the fastest and easiest way to get you started on a new codebase. To see how the docs render in markdown you can check them here - https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/opencv-python/on_boarding.md

The generation is based on static analysis and on top of that LLMs, and so far we are supporting python.

We would love to hear your opinion of the generated diagrams further we are interested in your current process of onboarding new developers and if we can be useful by any means.

Any feedback is more than welcome!

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/Camera Calibration & 3D Reconstruction.md
+++ b/.codeboarding/Camera Calibration & 3D Reconstruction.md
@@ -1,0 +1,60 @@
+```mermaid
+graph LR
+    OpenCV_Python_Wrapper["OpenCV Python Wrapper"]
+    Build_and_Setup_System["Build and Setup System"]
+    Utility_Scripts["Utility Scripts"]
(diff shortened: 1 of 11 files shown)


ivanmilevtues:
Hey all, I wanted to update you, than now we are also open-source so if you are interested you can take a look: https://github.com/CodeBoarding/CodeBoarding

Also I don't have access to add reviewers and it seems like the PR is being stale, anything I can do on my side for that?

PR #1107, opened 2025-06-12
Update build_wheels_windows.yml

Ryangx100:


asmorkalov:
The CUDA-enabled are tightly coupled with particular cuda version and has very large size. We cannot distribute it now.

PR #1121, opened 2025-07-07
Relax the build dependency on NumPy

mgorny:
Do not pin NumPy build dependencies to specific versions.  It is no longer necessary since NumPy 1.25, as it automatically exposes "an API that is backwards compatible with the oldest NumPy version that supports the currently oldest compatible Python version" -- effectively removing the need to maintain downstream compatibility matrices.

See: https://numpy.org/devdocs/dev/depending_on_numpy.html#adding-a-dependency-on-numpy

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,8 +1,9 @@
 [build-system]
 requires = [
   "numpy<2.0; python_version<'3.9'",
-  "numpy==2.0.2; python_version>='3.9' and python_version<'3.13'",
-  "numpy==2.1.3; python_version=='3.13'",
+  # NumPy >=1.25 defaults to exposing "an API that is backwards compatible with the oldest NumPy version that supports the currently oldest compatible Python version"
+  # https://numpy.org/devdocs/dev/depending_on_numpy.html#adding-a-dependency-on-numpy
+  "numpy>=2.0; python_version>='3.9'",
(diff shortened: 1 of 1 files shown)


asmorkalov:
Hello @mgorny Thanks a for the contribution. The versions are pinned for now on purpose: numpy switched to manylinux_x_y build environment since version 2.3. OpenCV still uses manylinux2014. We plan to revise dependencies and build environment for the next release and update build requirements.

mgorny:
Well, sure — but even then, using any `numpy<2.3` to build should be fine, right?

asmorkalov:
In general - yes.

Avasam:
This PR fixes being able to use dev numpy 3.15 wheels to build opencv-python from sdist, rather than having to build both.
This would also close:
- https://github.com/opencv/opencv-python/pull/1223
- https://github.com/opencv/opencv-python/pull/1170
- https://github.com/opencv/opencv-python/issues/1201
- https://github.com/opencv/opencv-python/issues/1165

Avasam (approved):

Avasam:
@asmorkalov 
> We plan to revise dependencies and build environment for the next release and update build requirements.

The "next release" (4.13.0.90) is long past. Can the build requirements be revised ?

Issue #1099, opened 2025-05-13
Musllinux wheels


fdelu:
Currently opencv wheels are only provided for the `manylinux` platform tags, for Linux distros that use the GNU C standard library. [PEP 656](https://peps.python.org/pep-0656/) proposes doing the same for with `musllinux` platform tags for distros that use the MUSL C standard library instead, for other distros such as Alpine Linux. This would enable much faster builds for projects using the official python alpine docker images. 

The [`manylinux`](https://github.com/pypa/manylinux) project repository provides base docker images for `musllinux` platforms as well, so that should ease the process.

PR #1133, opened 2025-07-31
Musllinux wheels  #1099 added

avdivan:


avdivan:
Tested build without GUI

Issue #1140, opened 2025-09-04
opencv-python-headless package relies on numpy below 2.3.0 ?


wyatt-wong:
I installeed the `pdf2docx` package but found that it relies on `opencv-python-headless` package and the `opencv-python-headless` relies on the older numpy package below 2.3.0 ?

Can `opencv-python-headless` package update to support numpy > 2.3.0 ? The latest numpy package is 2.3.2 as at 04-Sep-2025.

kvrban:
Same issue. I can not install picamera2 and opencv-python on my RPI 5 because of the issue.

Shiv-Expert2503:
**Testing Results: Python 3.10 vs 3.11 numpy Compatibility**

I've tested the numpy version conflict across Python environments and found the root cause:

**Python 3.10.18:**
- numpy 2.3.2: ❌ Cannot install (requires Python ≥3.11)
- Final result: numpy 2.2.6 + opencv-python-headless 4.12.0.88 ✅

<img width="1455" height="171" alt="Image" src="https://github.com/user-attachments/assets/2a81e273-ca29-4e11-a1e7-b270ea805965" />

<img width="341" height="258" alt="Image" src="https://github.com/user-attachments/assets/b2d21bbd-84a5-40da-bf7e-9127cad161ee" />

**Python 3.11:**
- numpy 2.3.2: ✅ Installs successfully initially
- **But when installing pdf2docx**: pip downgrades numpy 2.3.2 → 2.2.6 
- Final result: numpy 2.2.6 + opencv-python-headless 4.12.0.88 ✅

<img width="989" height="103" alt="Image" src="https://github.com/user-attachments/assets/84e29b2b-2249-4d7b-bfb7-0566066c6de2" />

<img width="1333" height="109" alt="Image" src="https://github.com/user-attachments/assets/958e260d-3114-4810-8030-9b7648f8a17c" />

<img width="300" height="279" alt="Image" src="https://github.com/user-attachments/assets/7dfd253d-38bb-4149-a9c4-4e2947a544ab" />

**Root Cause Identified:**
The conflict occurs because opencv-python-headless restricts `numpy<2.3.0`, while newer packages (like pdf2docx) work best with numpy ≥2.3.0. Pip resolves this by downgrading numpy.

**Proposed Solution:**
Update opencv-python-headless to use **conditional numpy requirements**:
- Python <3.11: `numpy>=1.17.2,<2.3.0` (current behavior)
[...]

Issue #1135, opened 2025-08-19
opencv-python 4.12.0.88 depends on numpy<2.3.0


Anna-Ku:
Hi,
**opencv-python 4.12.0.88** supports **numpy < 2.3.0**  and latest numpy version is 2.3.2.
Please update dependencies.
Thanks

Anna-Ku:
@asmorkalov

asmorkalov:
Numpy updated build environment from manylinux2014 to manylinux_x_y. The upper version is limited intentionally.

Houbein:
This is now also preventing the usage of Python 3.14 which needs the newest Numpy, please update!

Pranay-Pandey:
Can we please know the status of this? When is it likely to be released?

asmorkalov:
OpenCV team works on the new numpy support and switch to manylinux_2_28 in 4.13 (December).

pwuertz:
I don't quite understand.
Is there something OpenCV needs to work on for numpy 2.3 support, beyond not forbidding it in `setup.py`?
`opencv-python-headless==4.11.x` didn't define this upper constraint and works just fine with numpy 2.3.

mhicks-cfs:
Please update status.

Tenzer:
Version 4.13.0.90 was released yesterday which now depend on numpy >= 2.

PR #1142, opened 2025-09-11
allow numpy 2.3

braingram:
This PR relaxes the runtime numpy dependency to allow the latest released minor version (2.3.x).

~This follows the recommendations linked in https://github.com/opencv/opencv-python/pull/1119 (which added the runtime upper pin) and updates the pin based on the latest released version.~ EDIT: see https://github.com/opencv/opencv-python/pull/1142#discussion_r2352128587 for discussion on pin

issue #1140 #1135

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -22,11 +22,9 @@ def main():
     build_java = "ON" if get_build_env_var_by_name("java") else "OFF"
     build_rolling = get_build_env_var_by_name("rolling")
 
-    # NOTE: since 2.3.0 numpy upgraded from manylinux2014 to manylinux_2_28
-    # see https://numpy.org/doc/stable/release/2.3.0-notes.html#numpy-2-3-0-release-notes
     install_requires = [
         'numpy<2.0; python_version<"3.9"',
-        'numpy(>=2, <2.3.0); python_version>="3.9"',
+        'numpy>=2; python_version>="3.9"',
     ]
 
     python_version = cmaker.CMaker.get_python_version()


setup.py
```diff
@@ -26,7 +26,7 @@ def main():
     # see https://numpy.org/doc/stable/release/2.3.0-notes.html#numpy-2-3-0-release-notes
     install_requires = [
         'numpy<2.0; python_version<"3.9"',
-        'numpy(>=2, <2.3.0); python_version>="3.9"',
+        'numpy(>=2, <2.4.0); python_version>="3.9"',
```
opencv-alalek: 1) just `numpy>=2`
2) drop comment above this section

We should not push unnecessary locks on user environment.

@asmorkalov CI environment should be pinned instead (if necessary - probably not).

braingram: Thanks! Updated to `>2` in 7bd1825

vfazio:
This would be good to have considering python 3.14 is around the corner and numpy 2.3.x adds support for that version. 

Currently, opencv users cannot test against python 3.14 (easily).

I found this out today when trying to setup some 3.14 canary pipelines

opencv-alalek (approved):

vfazio:
Just saw this was approved, hopefully this gets merged and included in a release soon so we can derisk python3.14

vfazio:
Checking in on the status of this. 3.14 was released a few days ago, so there will be increased interest in getting support for a newer numpy and updated wheels for this repo

faverl:
> Revisando el estado de esto. La versión 3.14 se lanzó hace unos días, por lo que habrá un mayor interés en obtener soporte para un numpy más nuevo y ruedas actualizadas para este repositorio.

https://github.com/opencv/opencv-python/issues/1135#issuecomment-3448473109

altaic:
Is this planned to land before the next release? It'd be great to be able to support Python 3.14.

vfazio:
> Is this planned to land before the next release? It'd be great to be able to support Python 3.14.

Last thing I see related is https://github.com/opencv/opencv-python/issues/1135#issuecomment-3448473109

PR #1167, opened 2025-12-17
Use static BLAS and LAPACK dependnecy to reduce conflicts with other packages

asmorkalov:


rgommers:
Hi, I just happened to see this PR while looking at something else - just commenting in case it's useful. I do a lot of the maintenance of NumPy/SciPy wheels (and distro package issues), so if a conflict with NumPy or SciPy is the motivation here, happy to have a closer look.

OpenBLAS static builds aren't well-supported, so I suspect this is a dead end. Happy to be proven wrong of course, but if there's say a symbol conflict then better symbol hiding is probably a more promising avenue to explore.

Issue #1165, opened 2025-12-14
Dependency specification missing for python3.14


giochanturia:
There is no case defined for `numpy` using `python3.14` in `pyproject.toml`. As a result, installing packages that depend on `opencv-python` and a newer version of `numpy` does not work as expected and causes conflicts between `numpy` versions.

https://github.com/opencv/opencv-python/blob/81b4c07a3a29bdc6df70da84358b63ba1d656aae/pyproject.toml#L5

lucalista:
I also reported an issue related to a problematic numpy version (#1182) that was fixed with a manual upgrade of numpy version.

MDSALMANSHAMS:
This one, #1155 and #1160 are all the same root cause, and it looks resolved since 4.13.0.90.

The opencv wheel was never the problem on 3.14 — the wheels are `cp37-abi3`, so a 3.14 interpreter picks them up fine (both logs in #1155/#1160 show `opencv_python-4.12.0.88-cp37-abi3-win_amd64.whl` downloading successfully). What failed was the **numpy** requirement it pulled in. 4.12.0.88 capped it:

```
4.12.0.88   numpy<2.0; python_version < "3.9"
            numpy<2.3.0,>=2; python_version >= "3.9"     <- the cap
```

On 3.14 that cap resolves to numpy 2.2.6, which has no cp314 wheel, so pip fell back to building numpy from source — which is exactly the `numpy-2.2.6.tar.gz` → `Preparing metadata (pyproject.toml) ... error` both reporters hit.

The cap is gone from 4.13.0.90 onward (checked the PyPI JSON metadata for each release):

```
4.13.0.90   numpy>=2; python_version >= "3.9"
4.13.0.92   numpy>=2; python_version >= "3.9"
4.14.0.94   numpy>=2; python_version >= "3.9"
5.0.0.93    numpy>=2; python_version >= "3.9"
```

So pip on 3.14 now resolves to a numpy that actually ships cp314 wheels, and no source build is attempted.

For the specific ask in this issue — a `python_version` case for 3.14 in the build requirements — `pyproject.toml` on `4.x` now has one:

```toml
"numpy==2.1.3; python_version=='3.13'",
"numpy==2.3.2; python_version=='3.14'",
```

@asmorkalov this looks safe to close, along with #1155 and #1160, unless someone can still reproduce it on 4.14.0.94 or 5.0.0.93.

PR #1170, opened 2025-12-30
Add NumPy upper-bound pin for Python 3.14 compatibility

FaryalRizwaan:
This PR adds a Python 3.14–specific NumPy version constraint in pyproject.toml to fix dependency resolution failures when installing opencv-python on Python 3.14.

**Issue**: #1165 – Installation fails on Python 3.14 because no compatible NumPy version is selected during build dependency resolution.

### **Changes**

Added an environment marker in pyproject.toml to require numpy>=2.1.3,<2.2 for python_version >= '3.14'.

Preserves existing dependency behavior for Python 3.8–3.13.

No changes to setup.py; the fix is fully compliant with PEP 518 and modern packaging standards.

### Testing

**Verified dependency resolution using pip install --dry-run -e ..**

Confirmed that the new marker does not affect installation on supported Python versions (3.8–3.13).

The pinned NumPy version (2.1.3) is published on PyPI and declares Python 3.14 support via classifiers.

This minimal, forward-compatible change ensures opencv-python remains installable as Python 3.14 adoption begins.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -2,7 +2,7 @@
 requires = [
   "numpy<2.0; python_version<'3.9'",
   "numpy==2.0.2; python_version>='3.9' and python_version<'3.13'",
-  "numpy==2.1.3; python_version=='3.13'",
+  "numpy==2.1.3; python_version=='3.14'",
   "packaging",
   "pip",
   "scikit-build>=0.14.0",


ayushrajput6128:
This PR correctly adds a Python 3.14–specific NumPy constraint and resolves the missing dependency case described in #1165.
The environment marker is forward-compatible and preserves existing behavior for Python 3.8–3.13.
Looks good to me.

FaryalRizwaan:
Thanks for the review, @ayushrajput6128 ! Glad to hear the environment markers look correct to you.

mhicks-cfs:
Versions should not be overly constrained, please see #1140

lucalista:
Release [4.13.0.90](https://github.com/opencv/opencv-python/releases/tag/90), which supports Python 3.14, is not yet in PyPI. Is there any scheduled date for releasing it under PyPI?

rgommers (changes requested):
This PR is definitely incorrect. Note that exact pins for `numpy` are no longer needed since numpy 1.25.0 (see https://numpy.org/devdocs/release/1.25.0-notes.html#compiling-against-the-numpy-c-api-is-now-backwards-compatible-by-default).

A more reasonable change, which would be an actual improvement, would be to change "==2.1.3"` to `">=2.1.3"`.

rgommers:
And that approach was already taken in gh-1121 (still open), so I'd just close this PR.

Avasam:
Yeah this should be closed in favor of https://github.com/opencv/opencv-python/pull/1121

Mukller:
Heads-up: this and #1223 touch complementary halves of the same compatibility problem — this one pins the **build-time** numpy in `pyproject.toml`, #1223 adds **runtime** floors in `setup.py` `install_requires`. Worth coordinating so they land together rather than conflicting.

More importantly, I think the edit here moves the pin to the wrong place:

- `numpy==2.1.3` does not support Python 3.14 (cp314 wheels / compatible C-API arrived in the numpy 2.3.x line), so a 3.14 build isolated environment would either fail to resolve or silently build against an unsupported numpy.
- Meanwhile Python 3.13 is left with **no build pin at all**, losing the deterministic builds the exact-pin scheme exists for.

If the goal is 3.14 readiness, the minimal correct change should be *additive*:

```toml
  "numpy==2.0.2; python_version>='3.9' and python_version<'3.13'",
  "numpy==2.1.3; python_version=='3.13'",
+ "numpy==2.3.4; python_version>='3.14'",
```

(keeping the 3.13 line intact). The `2.3.x` micro-version should be whatever the project standardises on for cp314 builds.

Issue #1138, opened 2025-08-28
opencv-python doesn't have GUI ?


areche:
Hi! I am trying to install the GUI version of `opencv-python` as opposed to the non-GUI version called `opencv-python-headless`. However, no matter which version of opencv-python I install, the GUI is not always present. I had to go back to version `4.5.5.64` to find a wheel that has some sort of GUI.

```
# pip install opencv-python==4.5.5.64; python -c "import cv2; print(cv2.getBuildInformation());"
Collecting opencv-python==4.5.5.64
  Downloading opencv_python-4.5.5.64-cp36-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (18 kB)
Requirement already satisfied: numpy>=1.21.2 in /usr/local/lib/python3.12/dist-packages (from opencv-python==4.5.5.64) (1.26.4)
Downloading opencv_python-4.5.5.64-cp36-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (60.5 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 60.5/60.5 MB 52.6 MB/s eta 0:00:00
Installing collected packages: opencv-python
Successfully installed opencv-python-4.5.5.64
.
.
.
  GUI:                           QT5
    QT:                          YES (ver 5.15.0 )
      QT OpenGL support:         NO
    GTK+:                        NO
    VTK support:                 NO
.
.
.
# pip install opencv-python==4.6.0.66; python -c "import cv2; print(cv2.getBuildInformation());"
Collecting opencv-python==4.6.0.66
  Downloading opencv_python-4.6.0.66-cp36-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl.metadata (18 kB)
Requirement already satisfied: numpy>=1.21.2 in /usr/local/lib/python3.12/dist-packages (from opencv-python==4.6.0.66) (1.26.4)
Downloading opencv_python-4.6.0.66-cp36-abi3-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (60.9 MB)
   ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 60.9/60.9 MB 73.4 MB/s eta 0:00:00
Installing collected packages: opencv-python
Successfully installed opencv-python-4.6.0.66
.
.
.
  GUI:                           NONE
    GTK+:                        NO
[...]

PR #1181, opened 2026-01-15
solved #1138 by Adding missing Windows GStreamer build documentation

ayushrajput6128:
This PR fixes issue #1138 by adding the missing Windows GStreamer build instructions 
to the opencv-python repository documentation.  

Changes included:  
- Added step-by-step instructions for setting up GStreamer on Windows  
- Updated any relevant references and links  
- Ensured formatting matches existing documentation style  

This should help Windows users successfully build OpenCV with GStreamer support.

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,6 +161,38 @@ If some dependency is not enabled in the pre-built wheels, you can also run the 
     - Optional: on Linux use some of the `manylinux` images as a build hosts if maximum portability is needed and run `auditwheel` for the wheel after build
     - Optional: on macOS use ``delocate`` (same as ``auditwheel`` but for macOS) for better portability
 
+#### Building with GStreamer on Windows
+
+To build opencv-python with GStreamer support on Windows, you must install
+the official GStreamer MSVC binaries and make them available to CMake during
+the build process.
+
+**Prerequisites**
+- Windows 10 or later (64-bit)
+- Visual Studio 2019 or newer (MSVC)
+- CMake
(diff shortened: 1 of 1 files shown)


Mukller:
The section fills a real gap (#1138 — Windows GStreamer docs were indeed missing), but as written it can't be followed to completion:

1. **The actual build command is missing.** The *Build* section says "Run the standard manual build command:" and then ends — there's no command after the colon. Since this is the entire point of the section, it needs the concrete invocation (the `python setup.py bdist_wheel` / `pip wheel . --verbose` line Windows builders actually run, plus which shell it assumes).

2. **The environment-variable snippet isn't fenced.** Everywhere else in this README shell snippets use triple-backtick blocks; here `set PATH=...` / `set GST_PLUGIN_PATH=...` sit as bare text, which breaks rendering and copy-paste. They're also `cmd.exe`-style (`set X=Y`) — worth saying explicitly that these go into *cmd.exe*, not PowerShell (`$env:PATH = ...` there).

3. Consider mentioning `GST_PLUGIN_SYSTEM_PATH` alongside `GST_PLUGIN_PATH` — for runtime plugin discovery during tests of the built wheel, the `SYSTEM_PATH` variant is what GStreamer actually consults when `GST_PLUGIN_PATH` is meant for development plugins only.

With the command restored and the block fenced this is a clear keep — the prerequisites list itself (runtime + development MSVC installers, matching Python) is accurate.

PR #1190, opened 2026-01-24
manylinux: avoid bundling OpenSSL to fix FIPS import crash, Related issue: #28456

AdityaMishra3000:
### Problem

On FIPS-enabled systems, importing `cv2` aborts the Python process with:-

    OpenSSL FIPS SELFTEST FAILURE
    Aborted (core dumped)

I reproduced this for both `opencv-python` and `opencv-python-headless`.

### How I reproduced it

I reproduced this on Rocky Linux  with FIPS mode enabled
(`/proc/sys/crypto/fips_enabled = 1`) using the published Linux wheels:

    python -c "import cv2"

The process aborts immediately during module import.

Inspecting the wheel shows that `cv2.abi3.so` is linked against a bundled
OpenSSL (`opencv_python.libs/libssl*.so`), which is loaded at import time and
fails OpenSSL self-tests in FIPS mode.

### Root cause

The manylinux build currently builds and bundles its own OpenSSL and links
FFmpeg against it. On FIPS-enabled systems, this non-FIPS OpenSSL triggers
a self-test failure during `dlopen`, aborting the process before any OpenCV
code runs.

### Solution

This change removes the vendored OpenSSL from the manylinux build and relies
on the system OpenSSL instead. FFmpeg is still built with OpenSSL support, but
discovery is done via system pkg-config paths.

System OpenSSL is FIPS-compliant on FIPS-enabled systems, which prevents the
abort while preserving existing functionality.

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,3 @@
+* @skvark @alalek
+
+
--- a/docker/manylinux2014/Dockerfile_x86_64
+++ b/docker/manylinux2014/Dockerfile_x86_64
@@ -9,7 +9,6 @@ ARG FREETYPE_VERSION=2.14.1
 ARG LIBPNG_VERSION=1.6.53
 ARG VPX_VERSION=v1.15.2
 ARG NASM_VERSION=2.15.04
-ARG OPENSSL_VERSION=1_1_1w
 ARG QT_VERSION=5.15.18
 ARG YASM_VERSION=1.3.0
 ARG AOM_VERSION=v3.13.1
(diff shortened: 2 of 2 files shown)


AdityaMishra3000:
Related issue: https://github.com/opencv/opencv-python/issues/1191

AdityaMishra3000:
Requesting review from maintainers when convenient.

andznui (approved):

AdityaMishra3000:
Hey @andznui, thanks for reviewing the PR, i would appreciate a approving review inorder to merge this, also let me know if i need to make any changes here. 

Thank you

c-schoenherr:
Hello @AdityaMishra3000 and @andznui, is this branch still under review?

kohtala:
Will this work when OpenCV is run in a container running other distribution from the host?

asmorkalov:
OpenSSL removal is not an option for now as it's required for https URLs handling and RTSP cameras. I added detailed issue analysis to the mentioned issue. Let's continue discussion their as final decision has not been made.

marcreichman-pfi:
We ended up just building opencv from our own image baseline, including python, to get around this.

PR #1210, opened 2026-02-18
Add Python 3.14 to macOS ARM64 build matrix

### LINKED ISSUE #1196: No wheel available for OSX ARM for opencv-python==4.13.0.90
**Error:**
`error: Distribution opencv-python-headless==4.13.0.90 @ registry+https://pypi.org/simple can't be installed because it doesn't have a source distribution or wheel for the current platform`

**Reproduce:**
Run `uv add opencv-python-headless==4.13.0.90`

**Root cause:**
No compatible wheel exists for ARM OSX + Python 3.14.2

**Platform:**
- Apple M4 Max
- Python 3.14.2
- Package Manager: uv 0.9.27

shrutisachan08:
This PR add python 3.14 to macOS ARM64 build matrix.
Currently macOS ARM64 build wheels only using python 3.9 while the test matrix already include python 3.14.

Fixes #1196

### DIFF
--- a/.github/workflows/build_wheels_macos_m1.yml
+++ b/.github/workflows/build_wheels_macos_m1.yml
@@ -22,7 +22,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.9']
+        python-version: ['3.9','3.10','3.11','3.12','3.13','3.14']
         platform: [x64]
         with_contrib: [0, 1]
         without_gui: [0, 1]


loscrossos:
i think no PR is needed: the workflows builds an ABI wheel once with python3.9 that is compatible across all pythons. 

You can see this is the "test" arm that uses the "build" arm and tests it across python 3.9-3.14:

https://github.com/opencv/opencv-python/blob/4ddfc013fd1f13d9b9e379dbebf2cdbeb052e7f8/.github/workflows/build_wheels_macos_m1.yml#L65-L71

Avasam:
Yeah seems like this'll just slow down the builds for no reason.

PR #1223, opened 2026-05-09
fix: add explicit NumPy version constraints for Python 3.13+ and 3.14 (NumPy 2.x compatibility)

### LINKED ISSUE #1201: Poor compatibility with Python 3.14+NumPy 2.3.3
This issue was raised last year, but the version involved was Py 3.10 and 3.11, so an additional issue is being proposed for compatibility with the new version.

```python
Python 3.14.0 (tags/v3.14.0:ebf955d, Oct  7 2025, 10:15:03) [MSC v.1944 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> import cv2

A module that was compiled using NumPy 1.x cannot be run in
NumPy 2.3.3 as it may crash. To support both 1.x and 2.x
versions of NumPy, modules must be compiled with NumPy 2.0.
Some module may need to rebuild instead e.g. with 'pybind11>=2.12'.

If you are a user of the module, the easiest solution will be to
downgrade to 'numpy<2' or try to upgrade the affected module.
We expect that some modules will need time to support NumPy 2.

Traceback (most recent call last):  File "R:\Python314\Lib\_pyrepl\main.py", line 58, in interactive_console
    run_multiline_interactive_console(console)
  File "R:\Python314\Lib\_pyrepl\simple_interact.py", line 151, in run_multiline_interactive_console
    more = console.push(_strip_final_indent(statement), filename=input_name, _symbol="single")  # type: ignore[call-arg]
  File "R:\Python314\Lib\code.py", line 324, in push
    more = self.runsource(source, filename, symbol=_symbol)
  File "R:\Python314\Lib\_pyrepl\console.py", line 226, in runsource
    result = self.runcode(code)
  File "R:\Python314\Lib\_pyrepl\console.py", line 181, in runcode
    exec(code, self.locals)
  File "<python-input-0>", line 1, in <module>
    import cv2
  File "W:\cv2\__init__.py", line 181, in <module>
    bootstrap()
  File "W:\cv2\__init__.py", line 153, in bootstrap
    native_module = importlib.import_module("cv2")
  File "R:\Python314\Lib\importlib\__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
AttributeError: _ARRAY_API not found
Traceback (most recent call last):
  File "<python-input-0>", line 1, in <module>
    import cv2
  File "W:\cv2\__init__.py", line 181, in <module>
    bootstrap()
    ~~~~~~~~~^^
  File "W:\cv2\__init__.py", line 153, in bootstrap
    native_module = importlib.import_module("cv2")
  File "R:\Python314\Lib\importlib\__init__.py", line 88, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
ImportError: numpy.core.multiarray failed to import
>>>
```

after downgraded to np1.26.4:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
opencv-contrib-python 4.13.0.90 requires numpy>=2; python_version >= "3.9", but you have numpy 1.26.4 which is incompatible.
opencv-python 4.13.0.90 requires numpy>=2; python_version >= "3.9", but you have numpy 1.26.4 which is incompatible.
```
（note：it requires numpy2 **while installation**, but numpy 1 **during runtime**）
（WHAT THE F**K IS THIS???

mulhamfetna:
## Summary
- Fix NumPy 2.x ABI compatibility issue for Python 3.13+ and 3.14
- Changed  to specify explicit minimum NumPy versions per Python version
- This resolves the runtime error where wheels compiled against NumPy 1.x fail when used with NumPy 2.x

## Changes
- Python 3.9-3.12: 
- Python 3.13: 
- Python 3.14+: 

## Root Cause
The wheel metadata declared  as dependency at install time, but the compiled wheels may have been built against NumPy 1.x headers, causing ABI incompatibility at runtime.

Fixes #1201

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -24,9 +24,9 @@ def main():
 
     install_requires = [
         'numpy<2.0; python_version<"3.9"',
-        'numpy>=2.0.2; python_version>="3.9" and python_version<"3.13"',
-        'numpy>=2.1.3; python_version>="3.13" and python_version<"3.14"',
-        'numpy>=2.3.0; python_version>="3.14"',
+        'numpy>=2.0; python_version>="3.9" and python_version<"3.13"',
+        'numpy>=2.1; python_version>="3.13" and python_version<"3.14"',
+        'numpy>=2.3.2; python_version>="3.14"',
     ]
 
     python_version = cmaker.CMaker.get_python_version()
--- a/setup.py
+++ b/setup.py
@@ -24,7 +24,9 @@ def main():
 
     install_requires = [
         'numpy<2.0; python_version<"3.9"',
-        'numpy>=2; python_version>="3.9"',
+        'numpy>=2.0.2; python_version>="3.9" and python_version<"3.13"',
+        'numpy>=2.1.3; python_version>="3.13" and python_version<"3.14"',
+        'numpy>=2.3.0; python_version>="3.14"',
     ]
 
     python_version = cmaker.CMaker.get_python_version()


setup.py
```diff
@@ -24,7 +24,9 @@ def main():
 
     install_requires = [
         'numpy<2.0; python_version<"3.9"',
-        'numpy>=2; python_version>="3.9"',
+        'numpy>=2.0.2; python_version>="3.9" and python_version<"3.13"',
+        'numpy>=2.1.3; python_version>="3.13" and python_version<"3.14"',
+        'numpy>=2.3.0; python_version>="3.14"',
```
Avasam, line 29: Why the overly restrictive patch version pin ?

```suggestion
        'numpy>=2.0; python_version>="3.9" and python_version<"3.13"',
        'numpy>=2.1; python_version>="3.13" and python_version<"3.14"',
        'numpy>=2.3; python_version>="3.14"',
```

`python_version<` should also be redundant, given that on higher python versions, the numpy version is already restricted to a higher bound.

Unlocking the pin was also handled in https://github.com/opencv/opencv-python/commit/7bd1825cef8d82d86bc2f3f62055a64dc7a1a370

mulhamfetna: Good catch, thanks — the patch-level pins weren't deliberate, just over-specific. Applied your suggestion.

One small deviation: I kept a patch-level floor for 3.14. Checking PyPI, 2.3.0 and 2.3.1 ship no cp314 wheels — 2.3.2 is the first release that does — so `>=2.3.2` is the accurate floor there. `>=2.1` for 3.13 matches the same way, since 2.1.0 is the first with cp313 wheels.

Happy to drop the `python_version<` upper bounds as you suggested if you'd prefer — they are redundant, since the highest applicable floor wins. I left them only because your snippet kept them; say the word and I'll simplify.

PR #1232, opened 2026-06-09
Added tests with Python 3.14.

asmorkalov:


MugundanMCW:
Hi @asmorkalov,

There appears to be an issue with the Windows x86 and x64 wheel builders. The MSBuild action on the windows-latest runner is picking up the latest Visual Studio 18 toolchain, which is causing the problem. I don't see the same issue with the Windows ARM64 build, as MSBuild is using Visual Studio 17 in that case.

Thanks.

PR #1237, opened 2026-06-21
Remove python word from the package name.

asmorkalov:


abhishek-gola:
@asmorkalov, Will it be a symlink now i.e. user can download using `pip install opencv` as well as `pip install opencv-python`?

